PR #89, opened 2015-12-10
Enable bi-directional POI merge tool

brianhatchl:
@sisskind @bwitham Will you guys comment on the changes and merge?

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -172,7 +172,7 @@ en:
     nothing: Nothing to redo.
   review:
     merge:
-      description: Merge review features.
+      description: Merge review features. Hold Ctrl to reverse.
     previous:
       description: Skip to previous review.
     next:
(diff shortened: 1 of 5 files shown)


bwitham:
Changes make sense to me at first glance.  Might be worth running the code against some real world datasets to see if there are any unintended side affects. @jchoi1267 and I frequently used Brazil and MapCruzin data during the 5496 branch tests, as they usually revealed the most problems in the code.

brianhatchl:
I did my development using the DC poi data.  Seemed to work as intended, thought I didn't encounter that issue you saw with the Brazil data where a poi that was merged was involved in 600 other reviews.

bwitham:
Cool.  DC POI is another good one.  The jury is still out on the 600 interconnected review Brazil thing....meaning it may or may not be revealing some other underlying problem in the code.

Issue #128, opened 2016-01-04
Clipped Dataset goes to incorrect folder


sisskind:
After clipping a dataset, it does not go to the selected folder.

sisskind:
@mjeffe98 Ready for test

mjeffe98:
Tested on latest develop branch and clip goes to the designated folder now.  It will be included as part of the smoke test doc as well.

PR #152, opened 2016-01-05
Clipped Dataset Save Issues

sisskind:
* Clipped dataset will take placeholder value if nothing new added
* Clipped dataset now going to correct folder destination


sisskind:
ngageoint/hootenanny-ui#128

Issue #122, opened 2016-01-04
Multiline Relation does not show "blue" feature


sisskind:
When in review mode and if review feature is multiline relation, the "blue" feature is not being highlighted as blue.  The click to zoom appears to be working when you click on the table.

jchoi1267:
This smells like all the members of child relation is not being loaded

jchoi1267:
This bug was caused by not implmenting code to handle relation members when highlighting feature.
Now we recursively highlight members

mjeffe98:
I was also able reproduce a similar issue with the attached MassNh.zip data. [MassNh.zip](https://files.zenhub.io/5693ec47547e1dfa3ce9169f)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)
 
Review in question involves a relation (red) against a linear river feature (blue).  In this case, only the non-relation feature (blue) is highlighted even though both appear in the table.  Clicking on the red feature selects the relation but it does not get highlighted in the map.

mjeffe98:
Can @sisskind or @jchoi1267 provide the dataset that reproduced the issue?  Need to verify against that.

mjeffe98:
Not occurring with a subset of Haiti roads data.  "Blue" feature always appearing.

PR #155, opened 2016-01-08
Multiline Relation does not show "blue" feature #122

jchoi1267:
Was caused by not having code to hightligh relation members


jchoi1267:
Implemented code to highlight member features recursively

Issue #167, opened 2016-01-11
Blue highlighted review features not displaying at default zoom level for longer linear features


mjeffe98:
I started observing this on the release branch after a recent commit that impacted reviews for single ways against multilinestrings.  As I stepped through reviews of Haiti PortAuPrince data, noticed that on more than one instance the highlighted blue review against feature would not always display until I zoomed in further.  No longer zooming to the zoom level where both the highlighted features display.  It appears that you need to zoom in further to turn on the blue in this case.

@jchoi1267 @brianhatchl any ideas?  

![test](https://cloud.githubusercontent.com/assets/7560096/12250154/b3d6a944-b879-11e5-82c7-9228eb6e3f2b.gif)

mjeffe98:
This could be a duplicate of ngageoint/hootenanny-ui#122

mjeffe98:
reassigned to @brianhatchl since you are typically the color master :)

mjeffe98:
Adding some additional notes on this.  Rendering issues mainly occur on features > 2 km long.

mjeffe98:
Attached useful subset of Haiti data for recreating issue.  The  review against Rue Louis E. Pouget (red) vs. mgcp illustrates issue.  Only occurs when blue feature exceeds some unknown distance.  

[Haiti_subset.zip](https://files.zenhub.io/56953e195a0cfd1c5d6e86c8)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

brianhatchl:
@jchoi1267 Seems like loadMissingFeatureDependencies() is not handling relations (i.e. loading constituent ways of the relation)

jchoi1267:
As @brianhatchl  suggested fixed using Hoot.control.conflicts.actions.idgraphsynch.getRelationFeature

Also now delegates panning to caller since when zoomed out we do not want to zoom back in.

mjeffe98:
Looks good.  All segments were loading at the default zoom for each review.

Issue #144, opened 2016-01-04
Deleting line/area feature in relation leaves nodes


sisskind:
When you delete a line or area feature that is involved in a relation, it creates point features at each node that then need to be deleted as well.

mjeffe98:
Will investigate datasets to reproduce behavior.

mjeffe98:
@sisskind I cannot reproduce this error with any of the datasets that I have containing line or area features that are part of a relation.  In both cases I have deleted the feature containing the relation and no additional nodes have been added. Please supply the dataset where this has been reported.

sisskind:
![issue144](https://cloud.githubusercontent.com/assets/6191659/12241692/8f62dc16-b862-11e5-82eb-1d2aed0db211.gif)

mjeffe98:
@jchoi1267 @sisskind 
I was not able to recreate the exact same review scenario as one above in @sisskind comment but I was able to recreate with a different feature using a subset of the original source data (attached PortAuPrince_clipped.zip).  Road feature in conflated layer (Entree Desermithe) which is b/w the small town of Juvenot and the larger town Petionville shown here appears to be a way containing nodes that are part of a relation. The same road feature in the input datasets do not contain the same feature geometry and relation so this appears to be introduced during the merge. 

![image](https://cloud.githubusercontent.com/assets/7560096/12244382/4a1c3dd2-b858-11e5-8120-dffddcbf0278.png)

When you delete the feature in the merged output, there are remaining nodes as described in the ticket but these remaining nodes were in fact added during the conflation process and not part of the original input data sources.  Doesn't seem like a bug since these nodes are part of the relation and not part of the linear feature being deleted.

![relation_delete](https://cloud.githubusercontent.com/assets/7560096/12244585/49c9db54-b859-11e5-87fb-5ab683115102.gif)




 [PortAuPrince_clipped.zip](https://files.zenhub.io/56940480547e1dfa3ce919b4)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

mjeffe98:
This is just the screen grab of the POI markers that are generated in the conflated output for  the Entree Desermithe linear road feature.  Looking closely, it appears that they don't necessarily line up with the merged output feature but perhaps originate from either of the input layers.

![image](https://cloud.githubusercontent.com/assets/7560096/12245037/4f16c2f0-b85b-11e5-822b-865e1cca229b.png)

mjeffe98:
[...]

PR #170, opened 2016-01-13
Blue highlighted review features not displaying at default zoom level…

jchoi1267:
… for longer linear features #167

### DIFF
--- a/js/hoot/control/conflicts/actions/IdGraphSynch.js
+++ b/js/hoot/control/conflicts/actions/IdGraphSynch.js
@@ -236,7 +236,7 @@ Hoot.control.conflicts.actions.idgraphsynch = function (context)
             if(nMemCnt === 1){
                 _parent().actions.poimerge.disableMergeButton(true);
             }
-            _parent().map.featureNavigator.panToEntity(f, true);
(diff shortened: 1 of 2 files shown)


brianhatchl:
The blue review against feature is fixed, but I do see each vertex of the way being rendered, which I don't think we want.
screenshot from 2016-01-13 10 46 39
This may be due to disabling autoHidden feature filtering when in review mode:
https://github.com/ngageoint/hootenanny-ui/blob/develop/js/id/renderer/map.js#L144-L146

brianhatchl:
Hmm, that's not it.  But is seems the vertices should only be visible when the feature is selected.

mjeffe98:
@brianhatchl seeing the vertices maybe related to https://github.com/ngageoint/hootenanny-ui/issues/144.  It seems that the merged features that are green display vertices because those are nodes that have been added during conflation but pink/orange highlighed segments will not have the vertices pop-out like that.

brianhatchl:
The vertices disappear when you zoom in though.  And will reappear (though with a different symbol) when the feature is selected.

brianhatchl:
As this is most likely a symptom of loading members of a relation feature, like that described in #144, I'm going to go ahead and merge.

Issue #169, opened 2016-01-12
Clicking 'next' in fast succession produces error loading review members


mjeffe98:
Getting pop-up error message, "There are no members in the review relation" after going through a review of features containing multilinestrings vs. relations.  If you click the 'n' key in quick succession before the feature is loaded you will see this.  This may also be related to the NaN issue.  
(https://github.com/ngageoint/hootenanny-ui/issues/137) 

Not seeing an error in the dev console associated with this action.

![image](https://cloud.githubusercontent.com/assets/7560096/12275682/403f5d16-b926-11e5-8a33-d98223ce6a83.png)

mjeffe98:
Also was able to trigger NaN by clicking n really fast.

jchoi1267:
Updated the setProcessing masking. Although there is sometimes 0.5 sec gaps between locking it should cover most of processing range. Could not create lock covers whole process since tile loading is extremely asychronous operation and can not cover all entry and exit points

Issue #137, opened 2016-01-04
"Batman Mode"


sisskind:
Hootenanny will randomly go into Batman Mode...where coordinates are NaN NaN NaN NaN (get it?  Sing the Batman theme in your head...)

Screen goes black, appears hyper-zoomed in, and obviously coordinates are not a number.

mjeffe98:
Occurs mainly in review mode and possibly linked to certain reviews running on 0.2.19.  Retest with current version.

mjeffe98:
Will investigate to reproduce.

mjeffe98:
Recreated this with the PortAuPrince_osm/mgcp datasets (uncropped).  There are around 415 reviews in this data but I observed "Batman" mode when I clicked on a review against a relationand the coords switched to NaN NaN etc... and the map went black (URL coords also displayed as NaN,NaN etc..

mjeffe98:
example of what happens to the URL:  http://localhost:8080/hootenanny-id/#background=Bing&id=r111_550&map=NaN/NaN/NaN&overlays=,550

![image](https://cloud.githubusercontent.com/assets/7560096/12249843/58520750-b877-11e5-82bf-6dbb2ccdd14c.png)

mjeffe98:
@sisskind @jchoi1267 do we have enough information to assign to a developer?

jchoi1267:
Updated the setProcessing masking. Although there is sometimes 0.5 sec gaps between locking it should cover most of processing range. Could not create lock covers whole process since tile loading is extremely asychronous operation and can not cover all entry and exit points

mjeffe98:
Resolved > 100 reviews without getting the NaN issue to resurface.

PR #171, opened 2016-01-13
Clicking 'next' in fast succession produces error loading review memb…

jchoi1267:
…ers #169

 "Batman Mode" #137


mjeffe98:
hitting 'n' or 'r' at a typical user frequency does not trigger the error.  hitting 'n' key extremely rapidly will occasionally trigger the error but I believe this to be outside normal use.

Issue #224, opened 2016-02-10
Publish basemap button feedback


sisskind:
After clicking the publish basemap button, there should be some feedback (spinner, etc.) that the button has been pushed and the basemap is being loaded.  In version 2.20, nothing occurs to show the basemap is loading.

sisskind:
@mjeffe98 Ready for testing.

PR #239, opened 2016-02-15
On basemap publish, spinner appears

sisskind:
#224


mjeffe98:
@sisskind I noticed that the spinner only momentarily shows up in the Publish Basemap dialog and then it switches back to the original spinner in the Basemap tab.  Is this by design?
basemap_spinner

sisskind:
@mjeffe98 Yes.  The change here was the addition of the spinner on the "Publish" tab to show that the user clicked the button.  There were reports that a user would click and the window would hang as if nothing had happened, leading them to click multiple times.

Issue #203, opened 2016-02-02
Persist filter type in feature editing/creation


sisskind:
When creating a new feature, the filter by type should persist throughout feature creation (i.e. Select by type --> Edit Feature).

sisskind:
Needs to be tested -- current vagrant box is having a translation issue.

https://github.com/ngageoint/hootenanny-ui/commit/09f3ed7120029e696ce8be050cb8942ce651cc62

sisskind:
This is within entity_editor and preset_list.  Each has their own separate combobox, so there needs to be an interaction between the two in some manner.  Pushing to backlog until it becomes a higher priority.

sisskind:
@RDaviesDG put on backlog now since it was identified as not being a high priority, but feel free to tinker with if you have time.

mjeffe98:
I would make the case that this should be higher priority because it's part of the feature creation lifecycle.  If you are editing a layer and want to create a new feature, there is an expectation that the tag schema remain consistent with what has been selected in the Filter By Type drop-down.  If you create a new point, then select the point, select the tag schema e.g. TDS and then click on the feature type (Facility), the expectation is that the Filter by Type would continue to appear as TDS not OSM.  It requires far too many clicks for users to have to keep changing it.

mjeffe98:
Note that when you click on the feature subtype after the TDS tag schema selected in Filter by Type it switches back to OSM

![filterbytype](https://cloud.githubusercontent.com/assets/7560096/13235930/edeb8024-d975-11e5-8001-5aa76b482160.gif)

sisskind:
@jchoi1267 Can you review?

sisskind:
@jchoi1267 or @treycaldwell  Can you review?

PR #268, opened 2016-02-25
Added customTranslation service

sisskind:
* ngageoint/hootenanny-ui#203


sisskind:
@jchoi1267 Can you give this a once-over, then if good push to develop so @mjeffe98 can test?

Also - @mjeffe98 if this passes, should we merge into release (i.e. was this a regression test issue?)

mjeffe98:
@sisskind this was not identified during this sprints regression testing but last.  We could merge it into release after it passes.

jchoi1267:
@sisskind Looks good

Issue #356, opened 2016-04-08
Warn user when deleting dataset if bookmark exists


sisskind:
There have been issues when deleting a dataset if a current review bookmark exists (ngageoint/hootenanny#501).

Create a warning when a user is deleting a dataset that is affiliated with a current review bookmark.  Alert the user that if the dataset is deleted, bookmarks will be deleted as well.

mjeffe98:
@sisskind the behavior described in the ticket works.  I'll close it but bear in mind that users will still fall into trouble if https://github.com/ngageoint/hootenanny-ui/issues/377 happens.

PR #364, opened 2016-04-11
#356 Alert user that deleting layer will delete bookmarks

sisskind:
#356 

Also may mitigate ngageoint/hootenanny#501


sisskind:
@brianhatchl Can you review?  Should receive notification that the dataset has review bookmarks that will also be deleted.

brianhatchl:
Looking at the diff, I think it would be much more bulletproof if the delete map service call was updated to delete review bookmarks internally.  Here we are relying on the UI to call delete review bookmark before calling delete map.  Thoughts?

sisskind:
@brianhatchl do not disagree at all.  this was an initial band-aid from the UI side, but I think the best fix is what you described above.

brianhatchl:
K, I'll review this with that in mind and add another ticket to make the delete map service more robust.

Issue #404, opened 2016-04-28
Fix disabled mocha tests for iD and Hoot


brianhatchl:
Failing test specs are commented out in test/index.html.  When these are all passing, the tests on the js build output can be enabled for npm test in package.json

PR #405, opened 2016-04-28
fix and enable remaining mocha tests

brianhatchl:
refs #404


brianhatchl:
@sisskind Every mocha test is enabled except model/Import.js, so I'm going to merge to get this test coverage now.

Issue #408, opened 2016-04-29
eslint fix changing == to === will break badly formatted conditionals


brianhatchl:
I've seen uses of
`if (x === null)`
which will fail if x is undefined or anything not actually null

Replace with
`if (!x)`

brianhatchl:
Same goes for 
`if (x === true)`

brianhatchl:
Now I'm not sure about the true and false comparisons.  They may need to have explicit boolean type checking, so leaving them for now.

brianhatchl:
#410 merged

PR #410, opened 2016-04-29
fixes for #408

brianhatchl:


brianhatchl:
This fixes a js error that prevented shp/zip file types from being imported.

sisskind:
Looks good to me.

Issue #423, opened 2016-05-09
End of review error messages


sisskind:
Reaching the end of PaP conflation reviews. When going to delete a red feature, a neverending stream of red error notifications occurs eventually culminating the the complete and total meltdown of Chrome or Firefox and the world burns.

Error stream consists of:
TypeError: Cannot read property of id undefined.

There are no members in the review relation.

![image](https://cloud.githubusercontent.com/assets/6191659/15122270/464e82ae-15ec-11e6-9de5-30709a689a55.png)

sisskind:
https://github.com/DigitalGlobe/VGI-team-repo/issues/110

sisskind:
It appears that the code is getting stuck in a loop with LoadMissing, etc.  This does not occur when the features in the review are points/nodes.

sisskind:
This will occur no matter how many reviews are left.  Causation appears to be when all review features in relation are deleted.  This is ready for review in pull request.

sisskind:
Seems that this could be due to node relations not being deleted when they are empty, which is what occurs with way relations.  This should be fixed and then the code should be updated to work with empty relations with deleted features contained in history (as it currently stands, but needs to be updated when user resolves a review).

mjeffe98:
@sisskind there are 429 + reviews when you conflate the full PaP dataset but I've verified that you can delete all of the red/blue highlighted road segments in a review (including those that are part of a relation), click resolve and move onto next review.  I'm not entirely 100% certain what the conditions to fully recreate this are but it does appear that the previous errors associated with loading missing members of a relation have been resolved.

Issue #428, opened 2016-05-13
Empty point/node relations not deleting


sisskind:
Default iD behavior is to delete a relation when all members have been deleted.  Currently, relations made up of nodes are not deleted, but are left as modified.  This behavior should be reset to default iD behavior where an empty relation is deleted.

References:
- iD Editor issues 465, 1454
- https://github.com/openstreetmap/iD/commit/163c85bacbd254ee62153f6c7f6fbb09bf711213
- https://github.com/openstreetmap/iD/commit/163c85bacbd254ee62153f6c7f6fbb09bf711213#diff-0f5ef230bebc75bffbe8e23a5b67308cR51

sisskind:
https://github.com/ngageoint/hootenanny-ui/commit/cade0a90d3582831950ab79fb40a3b5ad9dcebe8

sisskind:
Will be handled in the #427 pull request

PR #436, opened 2016-05-17
#423

sisskind:
Please refer to https://github.com/ngageoint/hootenanny-ui/pull/427 for additional notes.

### DIFF
--- a/js/hoot/control/conflicts/actions/IdGraphSynch.js
+++ b/js/hoot/control/conflicts/actions/IdGraphSynch.js
@@ -51,7 +51,6 @@ Hoot.control.conflicts.actions.idgraphsynch = function (context)
         } else {
             // Check to see if the relation has been deleted.  If so, do not continue...
             if((_.find(context.history().changes().deleted,{id:fid}))){
-                _validateMemberCnt(fid,null);
                 context.hoot().control.conflicts.setProcessing(false);
                 return;
             }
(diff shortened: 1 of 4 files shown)


sisskind:
@brianhatchl new branch is 423-2.  423 will no longer be updated and will be deleted.  Created new branch to avoid detached head due to reverted code.

sisskind:
@sisskind -- note for self:  determine if there is a way to see if relation exists before running through LoadFromURL (@brianhatchl chime in if you are familiar with a method).  Currently looking through Java service code to determine if this method exists.  All I have found thus far is a call to load relations (http://localhost:8080/hoot-services/osm/api/0.6/relations?mapId=1036&elementIds=3). Ideally, I would like to use this to get all relation ids.  If successful, this will eliminate the warnings in the console that the website is not available.

brianhatchl:
I think a 404 for a request for a deleted relation isn't the worse thing.  That's the way REST is supposed to work.

sisskind:
Okay - if we are okay with the 404 appearing in console then this may be done. Will give one more pass to test.

sisskind:
@brianhatchl Think this is ready for review, with two caveats:

1. You will still see 404 errors in the console when saving an empty relation of ways.  This will be addressed in #439.
2. Empty point relations are not being deleted.  This will be addressed in #428.

How to test:
- Run conflation job with nodes as features
- Delete all red/blue features and get no error messages
- Click Resolve, get no error messages in UI
- Repeat with next review (delete...)
- Click Save, get no error messages in UI
- Repeat for ways

Issue #416, opened 2016-05-04
Undo merge does not change color back to source layer


sisskind:
When in review mode, if you undo a merge operation, the color of the point does not return back to the source layer.

brianhatchl:
Could be a complementary bug to this https://github.com/ngageoint/hootenanny/issues/658.

mjeffe98:
tested merge, then undo.  Looks like when merging 1 orange (layer1) with 1 pink (layer2), it takes two undo's to revert back to the original colors.  That seems expected since a merge is actually > 1 update to the data.

brianhatchl:
@sisskind had the same question and yes, a POI merge is an update tag event and a delete event, so two steps.

PR #438, opened 2016-05-18
let change tags update the feature, don't do it in javascript #416

brianhatchl:
refs  #416


sisskind:
@brianhatchl is anticipated behavior having user click undo twice?  

issue-416

brianhatchl:
It's always been like that.  The merge is a two-step process:

1. Update tags of destination feature.
2. Delete source feature.

So it takes two undos to revert it.

sisskind:
In that case good to go. Will merge and close.

Issue #424, opened 2016-05-09
Return of Batman Mode


sisskind:
I have occasionally been getting this issue when adding the a merged layer in Hoot and the result is black space with the exception of red and blue boxes containing the attributes.The coordinates say NAN_NAN_NAN_NA. I refresh Hoot and screen zooms out to the globe. Has this happened to other users and is there a fix?

![image](https://cloud.githubusercontent.com/assets/6191659/15122401/f8617bcc-15ec-11e6-8264-0fe9face8d03.png)

sisskind:
https://github.com/DigitalGlobe/VGI-team-repo/issues/115

sisskind:
Adding checks for map extent, center, zoom to ensure coordinates are accurate.  This should mitigate issue, and will mark closed at that point.  If it continues, new ticket will be opened with details and how to reproduce, if possible.

sisskind:
Using San Diego data, should get error message instead of going to NaN NaN

mjeffe98:
@sisskind I could only reproduce this error on selma:8093 and no where else with the same San Diego rivers data.  On selma got the NaN error below but on my local vm and on the new AWS I did not get the NaN so it may be difficult for me to reproduce.  

d3.min.js:1 Error: <path> attribute d: Expected number, "M0.5,0.5v8hNaNv-8".u @ d3.min.js:1
d3.min.js:1 Error: <text> attribute x: Expected length, "NaN".u @ d3.min.js:1

sisskind:
I think we should mark OBE.

PR #448, opened 2016-05-23
Check for valid coordinates

sisskind:
New function checks for valid coordinates when panning to entity.

@brianhatchl can you take a quick look -- should be an easy one.  Thanks.

This is a potential mitigation for #424, but will need to replicate the error to fully confirm.

### DIFF
--- a/js/hoot/Hoot.js
+++ b/js/hoot/Hoot.js
@@ -286,7 +286,7 @@ Hoot.hoot = function (context) {
 
             return true;
         } catch (err) { return false; }
-    }
+    };
 
 
     hoot.checkForSpecialChar = function(str){
--- a/js/hoot/Hoot.js
+++ b/js/hoot/Hoot.js
@@ -271,6 +271,24 @@ Hoot.hoot = function (context) {
         Hoot.control.TranslationAssistant(context);
     };
 
+    /**
+    * @desc Check for valid coordinate pair
+    * @param Coordinate pair [x,y]
+    **/
+    hoot.checkForValidCoordinates = function(coord){
+        try{
+            if(coord.length !==2) {return false;}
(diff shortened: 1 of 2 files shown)


brianhatchl:
It seems like several things could be going wrong that this PR is meant to mitigate.  

- entity extent not defined
- entity not in graph
- entity loc is invalid
- map extent not defined
- map extent is invalid
- map extent coords are invalid

Do we care what/why it is happening?  Or are we satisfied that when it does happen the code handles the error?

brianhatchl:
The San Diego conflated waterways datasets produce this error for me, but it's not "on command".  Some combination of resolving or skipping reviews will make it occur.

sisskind:
I will try with this dataset and see if I can get it to cause error.  Did error occur within branch or on current develop?

brianhatchl:
This branch.  The error was handled, but I saw the "Could not locate selected feature with id:" error alert.

sisskind:
Oh -- then that should be considered a win.  Point is to catch error and not go into NaN NaN.

sisskind:
Wait -- hold off on merging PR.  May have found root of issue.

sisskind:
It looks like some features have an extent of 'Infinity'.  The code above should mitigate that and throw the error.  Digging a touch deeper to see if the actual coordinates can be extracted.  If not, will keep this as is to mitigate the NaN issue.

sisskind:
@brianhatchl I think this is okay to go.  The error is going to catch any NaN values and will not allow iD to get "lost in space".  Is there a concern over the error message even if it does eventually get there?

brianhatchl:
Nope.  I will merge.

sisskind:
:)

Issue #444, opened 2016-05-20
Remove Detail section from About Hootenanny dialog


brianhatchl:
The detail section displays internal configuration not appropriate for the web UI

brianhatchl:
This also includes removing details section from Error Log (renamed Server Log) section of manage tab.

brianhatchl:
The export error log still has the hoot version information embedded as it's useful once the log is "detached" from the UI.  However, version detail for hoot-services has been omitted as described here, https://github.com/ngageoint/hootenanny/issues/747

PR #457, opened 2016-05-25
#444 remove detail section

brianhatchl:


brianhatchl:
Do we even need the About section in the Manage tab?

Issue #452, opened 2016-05-23
Incorrectly Rendering Map Layers


sisskind:
When you click the "related reviews" button on the pallette on a feature that is part of a RELATION review and no related reviews display, and then pan and zoom around, the map layer remains at the location and zoom level it was at when you clicked "related reviews".

sisskind:
https://github.com/DigitalGlobe/VGI-team-repo/issues/136

mjeffe98:
Now when you are able to click on a feature in a review b/w a relation and a way, it seems to display the bubble letters and allow users to pan, zoom, etc.  I have also observed the same behavior that @brianhatchl reported in https://github.com/ngageoint/hootenanny-ui/pull/460#issuecomment-221953594 where you cannot view related reviews on a feature that is part of a relation because the  relation is either not clickable to open the radial menu or it doesn't display the show related reviews icon.  When you click on the feature that is not a relation, this does have the radial menu with the show related review options and displays A-Z letters with lines pointing to each segment.

PR #460, opened 2016-05-26
#452, handle related reviews when one review feature is a relation

brianhatchl:


brianhatchl:
@sisskind ready for review.

The only caveats are that when related reviews are opened for a way vs a relation, the multiple parts of the relation each get their own label letter, even thought those letters all reference the same review.  I tried to get them to share a letter, but this broke the default behavior for a feature with multiple reviews.

And I noticed that for reviews of relation against relation, there is no way to open the related reviews because the relation is not clickable on the map and cannot open a radial menu.

Take a look, I used the Congo data with generates way vs relation reviews during conflation.

PR #488, opened 2016-06-09
Add MGCP to the list of presets

mattjdnv:
See Hoot #793 for the MGCP to/from english scripts


mattjdnv:
I screwed up and edited the branch in Master instead of Develop. I'll rebuild it and then re-do the pull request.

Issue #500, opened 2016-06-15
Error when loading a HGIS validation layer


brianhatchl:
Run through the HGIS workflow to create a validation layer. When you add it to the map the following error occurs:

Uncaught TypeError: validation.loadFeature is not a function

mjeffe98:
I'm getting the error message below on certain POIs in my test layer (
[Samoa_MGCP_Validation.zip](https://github.com/ngageoint/hootenanny-ui/files/319166/Samoa_MGCP_Validation.zip)
) making this had to confirm as fully functional.

![image](https://cloud.githubusercontent.com/assets/7560096/16130337/7de88248-33bd-11e6-9466-0abc80ca9be4.png)

mjeffe98:
It seems that the same POIs are getting loaded twice which doesn't look right as I would expect only tagging a feature once.  Also if you click the Save button it tries to commit a deleted relation.

brianhatchl:
There is another bug https://github.com/ngageoint/hootenanny/issues/846 that might be causing this.  This ticket was just to remove the error that occurred when loading a validation layer and going into validation mode.

mjeffe98:
Ah ok.  I won't conflate the two issues then.

PR #501, opened 2016-06-15
#500, multiple errors during hgis validation workflow

brianhatchl:


brianhatchl:
Hold off on this, the validation workflow isn't exiting cleanly.

brianhatchl:
And I see it failed eslint (whose changes broke things to begin with), so I'm on that too.

Issue #482, opened 2016-06-08
Export statistics to .csv


sisskind:
Provide the ability to export statistics to a .csv file

PR #507, opened 2016-06-16
#482, make the stats info downloadable

brianhatchl:
tab separated instead of comma separated though
mime-type should make it open in spreadsheet software


brianhatchl:
There is now a Download link:

screenshot from 2016-06-16 16 41 53

Which will download a .tsv file that can be parsed and opened by spreadsheet software:

screenshot from 2016-06-16 16 44 07

Issue #515, opened 2016-06-21
Limit how far the user can expand the conflation area


briandaviddavidson:
Currently, the dataset area can be expanded to cover the entire screen. This should be limited to at max, half the screen

Example:
![example](https://cloud.githubusercontent.com/assets/2423402/16240097/ef4e24dc-37b5-11e6-9804-00277c462450.png)

mikejeffe:
@briandaviddavidson did you do a submodule sync after merging?  I just rebuilt and there didn't appear to be any changes in hoot-ui. I can still expand to full screen.

briandaviddavidson:
No, I didn't. I'm not sure how to do that either. Is there something special I need to do in the commit?

sisskind:
@mikejeffe pull request hasn't been done yet.

mikejeffe:
Ok. I was just looking at the label change so I'll wait till the PR

PR #528, opened 2016-06-27
515

briandaviddavidson:
For ticket #515: Quick CSS fix to prevent users from expanding the left bar pass half of their screen


briandaviddavidson:
Made changes on a branch based off of `master` instead of `develop`. Going to recreate these to follow the gitflow

PR #529, opened 2016-06-27
#515: Add min and max width to sidebar

briandaviddavidson:
Created a feature branch off of `develop` instead of `master`

Ticket: #515

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -755,6 +755,8 @@ a:hover .icon.out-link   { background-position: -500px -14px;}
     overflow: hidden;
     z-index: 10;
     background: #f6f6f6;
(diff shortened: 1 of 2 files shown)


sisskind:
@briandaviddavidson think the sidebar should start at smallest width to save screen size.  Thoughts?

briandaviddavidson:
@sisskind It should still start at 400px. Does it not?

sisskind:
When I run 515 branch, it is defaulting to 664.44px.  Unsure if it is being overwritten by something else.

briandaviddavidson:
Give that a try. That should fix it.

sisskind:
Great job.

Issue #523, opened 2016-06-22
Expanding width of left frame needs improvement


sisskind:
The left frame is now expandable, but it doesn’t' evenly adjust the widths of the interior columns. Example, when you expand the left frame when looking at the statistics under the Options, if you expand the left frame it mostly effects the first column in the table and not the second.

briandaviddavidson:
Pull request has been submitted

PR #533, opened 2016-06-28
#523: Expanding width of left frame needs improvement

briandaviddavidson:
There was a max width that was preventing the inputs from expanding all the way, so that has been removed and matched to the sidebar code from #515 

Ticket: #523


sisskind:
test this please

sisskind:
@briandaviddavidson Take a look -- something screwy happening:

image

Issue #542, opened 2016-07-06
Fix issue of appending id when not available


briandaviddavidson:
In https://github.com/ngageoint/hootenanny-ui/issues/518, some of the code ([specifically this commit](https://github.com/ngageoint/hootenanny-ui/commit/c32a30f17a3e70a04562711e0c5366aa3deeb7cc)) appended an ID which broke other parts of the code. This needs to be corrected.

briandaviddavidson:
This was fixed in the PR above.

If `d.ImageryUsed()` is a string, an ID will be appended, otherwise it will pass over that attribute.

mikejeffe:
no need to keep this open.  If a developer can confirm than just close it.

briandaviddavidson:
@jcp2236 Can you please confirm and close.

PR has already been merged: https://github.com/ngageoint/hootenanny-ui/pull/543

jcp2236:
Confirmed.

PR #543, opened 2016-07-06
#542

briandaviddavidson:
This should fix the issue @sisskind ran into as discussed on Gitter

> Had to revert changes from ngageoint/hootenanny-ui@c32a30f, as it was causing an issue with the forceAddLayer function, which loads layers from review bookmarks. Was not sure which ticket it was related to, but it appears that not all 'd' variables have the imageryUsed function. When you go to fix it, test using the load review bookmark feature.

I added data and tested adding and loading a review bookmark and all seems to be in working order.

Ticket: #542

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -54,6 +54,7 @@ body {
 .limiter {
     position: relative;
     max-width: 1200px;
+    margin-left:10px;
 }
 
 .spinner {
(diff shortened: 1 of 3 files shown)


sisskind:
test this please

sisskind:
test this please

Issue #521, opened 2016-06-22
Expose OSM as export option


sisskind:
The user must choose a translation option, either tds or mgcp when exporting as osm. I had the default tds selected when I exported to osm although I don't think it translated anything based on reviewing the data in josm.

sisskind:
![image](https://cloud.githubusercontent.com/assets/6191659/16701170/0da8ed6c-452c-11e6-8c63-5f21b09cb1b8.png)

![image](https://cloud.githubusercontent.com/assets/6191659/16701305/e779b0c6-452c-11e6-9c2e-414d23fc5244.png)

PR #555, opened 2016-07-08
No Translation option for OSM Export Format

sisskind:
Also noticed that if 2 export dialogs are open, can override.  Added container variable to capture that in the rare occurrence it happens.

Issue #521

### DIFF
--- a/js/hoot/control/Export.js
+++ b/js/hoot/control/Export.js
@@ -147,6 +147,7 @@ Hoot.control.export = function (context, sidebar) {
                         
                         d3.select(this)
                         .on('change',function(){
+                            checkForOSM();
                             checkForTemplate();
                         });
                     }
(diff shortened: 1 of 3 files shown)


sisskind:
test this please

sisskind:
test this please

brianhatchl:
Shouldn't the option for no translation be available for all export formats?  e.g. If I want a shapefile but with OSM tags?

mikejeffe:
the slightly confusing aspect of this PR is that regardless of whichever target translation is specified, if the export format is set to .osm translation is ignored (see https://github.com/ngageoint/hootenanny/issues/863).  I can see how this makes the process a bit more transparent to the end user but it doesn't negate that does that precondition condition does it?

brianhatchl:
No, I would think you'd have to hide the translation drop down when OSM format is chosen.  However, from the renderdb export server I wrote I know hoot can export translated data to osm, so it could work, not sure why that combo is disallowed.

sisskind:
The requirement that came in was to offer a "No Translation" option specifically when exporting to OSM format.  @mikejeffe is correct that this is for increased clarity/transparency to end user.  @brianhatchl that is another option, if OSM format is selected to hide the combobox.  I will reach out to stakeholder to see what they prefer.

brianhatchl:
I thought there was another issue in that if a translation is selected for OSM format, it is not used.  If that is the case and we don't allow that ( I think we should ) then the translation option for OSM should ONLY be No Translation.

sisskind:
Closing this Pull Request to replace with future one as the scope of this issue has changed.

PR #557, opened 2016-07-11
hitting enter adds folder instead of refreshing

rmrice:
added keypress function to input for add folder (same code as onclick).

### DIFF
--- a/js/hoot/control/utilities/Folder.js
+++ b/js/hoot/control/utilities/Folder.js
@@ -519,7 +519,6 @@ Hoot.control.utilities.folder = function(context) {
 
     hoot_control_utilities_folder.importFolderContainer = function (data) {
         context.hoot().model.folders.listFolders(context.hoot().model.folders.getAvailFolders());
-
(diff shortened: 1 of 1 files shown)


sisskind:
test this please

sisskind:
@rmrice Challenge for you:  Instead of writing the "enter" submit and the "apply" button click functionality twice (since it is the same), could you write a function that captures it once and then call it from either event?  There are a few examples of this in the code I can point you to.

rmrice:
@sisskind sure, can you point me to the examples?

sisskind:
https://github.com/ngageoint/hootenanny-ui/blob/master/js/hoot/view/utilities/Dataset.js#L184

The _deleteLayerCallback is a variable that is a function, which can be called from within another function.  You could call the new function from within the button click and the enter key.

rmrice:
updated Folder.js to call a _submit function when enter key is triggered and on clicking the Add Folder button

sisskind:
test this please

Issue #508, opened 2016-06-16
Layers are remaining in sidebar


sisskind:
Appears that layers that are loaded are not removed from sidebar after they are loaded and removed. This may be a feature of iD editor.

![123](https://cloud.githubusercontent.com/assets/6191659/16134625/39755fe4-33ec-11e6-8b7a-96a7df1e0516.png)

sisskind:
https://github.com/DigitalGlobe/VGI-team-repo/issues/165

sisskind:
It is possible that we are inadvertently adding layers to the iD layer list and the Hoot layer list.

briandaviddavidson:
Looks like the layer needs to be there for the bottom layer list to render. Not sure why, but for now, I'm going to work on hiding the list item, then work on removing it completely.

Might be something that happens in the refactoring process.

briandaviddavidson:
Testing when looking at the PR revealed more issues that I will be looking into

briandaviddavidson:
Looks like this was relying on another PR that was not in the feature branch. After pulling in changes from `develop`, everything is operating as expected.

PR #559, opened 2016-07-11
508

briandaviddavidson:
Layers now are hidden from view in the `layer-list`, but users can still interact with them in the `hoot-layers`.

I also, added [this code](https://github.com/ngageoint/hootenanny-ui/commit/8a76ec3a0ab96ea39b0fb424f6fd5c4f64346ac4) to fix the arrows looking off center in the `hoot-layers` list. Can easily be changed back if it's not the desired look. 

Ticket: #508

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -2036,6 +2036,11 @@ div.full-screen > button:hover {
     text-overflow: ellipsis;
 }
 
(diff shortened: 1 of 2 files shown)


sisskind:
test this please

PR #561, opened 2016-07-13
545

rmrice:
Re-added code that triggers spinner icon when importing data. Commented out code that triggers progrses bar. Now when hitting "import" just the spinner should show up. (starting at line 220, ImportDataset.js)


briandaviddavidson:
Was this created off of `master` or `develop`? This seems like a lot of code for being created off of `develop`

sisskind:
@briandaviddavidson @rmrice looks like the PR is going to add into master as opposed to develop.  Best thing to do is to create a PR with branch and develop and close this one.

Issue #499, opened 2016-06-13
Search not working when MGCP is active schema


mikejeffe:
Seems that I can't search for a feature type when the MGCP filter by type option has been selected as a result of https://github.com/ngageoint/hootenanny-ui/issues/487.  Other searches work for osm, tds but not for mgcp.  I tested this by searching for known feature types for each schema e.g. osm: soccer*, tds: ford, etc...

@brianhatchl @sisskind

mikejeffe:
Doesn't look like the search is working with MGCP yet.  I created a feature, flipped the filter to MGCP and searched for bridge.  None of the search results had the name "bridge" in it.  Contrast this with OSM and TDS where search terms appear in result.

brianhatchl:
MGCP presets are now there (they weren't at all before), the search fix is being tracked here https://github.com/ngageoint/hootenanny/issues/907

mikejeffe:
thanks @brianhatchl.

mikejeffe:
this ticket was originally opened because search wasn't working with MGCP selected so I don't think I can close it based on whether the presets are there.  I'll track this with ngageoint/hootenanny#907.  Once that work is done, we can probably close this ticket.

mikejeffe:
closed after testing ngageoint/hootenanny#907

PR #563, opened 2016-07-14
fixes for #499

brianhatchl:
Add presets for MGCP and change schema filter behavior to only show presets for selected schema


brianhatchl:
@sisskind can you try this?  The search is returning results, but they looks like pretty poor matches for the search term used.

1. Add a new point to the map.
2. Choose the TDS or MGCP schema in the drop down.
3. Type something like "building" or "hospital" in the Search bar.

briandaviddavidson:
I went in, added a point, switched to TDSv40/TDSv61 and was able to search for building and have building be the first option, but I couldn't get the same from MGCP. Is that expected @brianhatchl?

cc @sisskind

sisskind:
Can confirm that something is awry with MGCP.  TDS works, but appears that MGCP is returning same results regardless of search term (Ferry Crossing Point Feature, Extraction Mine Point Feature, Quarry Point Feature...).

Upon further investigation, is it possible that MGCP is being searched against feature code and not description?

image

image


image

sisskind:
@briandaviddavidson @brianhatchl working off of hoot develop branch

sisskind:
...and after more discussion, it is possible that what is aforementioned is a bug that already existed and has nothing to do with this ticket.  If that is the case, we should look into creating a new ticket to solve issue.

brianhatchl:
Please merge, will move this issues described above into new ticket

Issue #556, opened 2016-07-08
Cannot double-click to complete new feature for line/area when line snapping disabled


mikejeffe:
This issue is not the easiest the reproduce but I'll try here and I have been able to duplicate it on AWS.  I noticed that more often then not in Chrome primarily, when creating a new line/area geometry I could NOT double-click to complete the feature, it would simply continue adding new nodes unless I clicked a key to escape this.   When I clicked any number key e.g. 1,2, 3, etc., the enable line snap pop-up appeared and the feature completed.  In other words, on my local dev ubuntu instance I could only complete new features if I repeated this or I checked the enable line snap check box in Map Data options.  

On AWS, I wasn't getting that issue (i.e. not being able to complete a new line/area by double-clicking) to occur by default but when I checked/unchecked line snapping it did appear to be an issue.  Again this is occurring primarily in Chrome.  I wouldn't be surprised if this was difficult to reproduce but I've noticed it enough to log a ticket for it.  Maybe someone from the UI team can reproduce: @sisskind @briandaviddavidson @jcp2236 @brianhatchl

sthorneloe:
@mikejeffe OK, nice to see this here, I thought it was just me (Hilary has no problems for some reason).

brianhatchl:
This bug is caused by disabling snapping in this ticket https://github.com/ngageoint/hootenanny-ui/issues/520

brianhatchl:
This chunk of code seems to have been customized by hoot, but I'm not sure we need it.
https://github.com/ngageoint/hootenanny-ui/blob/develop/js/id/behavior/draw.js#L77-L93

Making this chunk like default iD code fixes the double-click to finish drawing a new feature.  Not sure if it breaks any other behavior though.

PR #564, opened 2016-07-14
fix for #556, double-click should finish drawing feature

brianhatchl:


briandaviddavidson:
Confirmed working.

Double click now ends drawing of lines and areas.

Issue #449, opened 2016-05-23
Import should support zipped osm and gdb datasets (not just shapefiles)


brianhatchl:

treycaldwell:
When importing a Dataset there are two potential error conditions a user could encounter when attempting to upload an osm file that has been zipped (here, named sample.osm.zip).

1. User chooses (shp,zip) as Import Type, selects the file, and then encounters a UI error generated by ImportDataset.js indicating "Please select only single type of files. (i.e. can not mix zip with osm)"
2. User chooses (osm) as Import Type, selects file and clicks Import, and then encounters a server-side generated error (FileUploadResource.java) "Failed upload: Hootennany does not support zip files containing .osm data."

brianhatchl:
If the hoot command (using ogr) does not accept a zipped osm or fgdb file, then I suggest the service do the work of unzipping.  See how the OgrAttributesResource handles things.

treycaldwell:
Don't invoke the `makeetl` makefile from the command line without setting the appropriate environment variables or it will do a `rm -rf $HOOT_HOME`

bwitham:
nasty

bwitham:
I'd do all my debugging for this by using junit test drivers that invoke the java resource file which is actually calling that make script.  Less tedious imo.

treycaldwell:
Yeah, I'm not going to dwell on the inner workings of the JobControllerBase and JobExecutioner and why we're calling makefiles.  I'm just going to get in there, get it working, and get out as @brianhatchl said.  However, we should have a discussion about alternatives to calling the core without using makefiles, parsing STDOUT, etc.

bwitham:
There definitely may be ways to make it more that interface more straightforward I think.  All involve a lot of work.  The brick wall you'll always hit is that hoot core is largely not thread-safe.  Hoot lived its life completely on the command line a few years before it was service enabled, and at the time the services were created we realized the level of effort to make that code threadsafe was far beyond what we were willing to invest or were capable of investing given the much smaller team size at the time.  That decision was actually made a few years ago, so I can't imagine how much harder it would be to make all of it threadsafe now.  

[...]

PR #565, opened 2016-07-18
Issue #449: Import should support zipped osm and gdb datasets

treycaldwell:
Changes to import dialog to allow uploading a .osm.zip file
and to show in Chrome the option for choosing a .gdb.zip.


js/hoot/control/utilities/ImportDataset.js
```diff
@@ -454,7 +454,7 @@ Hoot.control.utilities.importdataset = function(context) {
         } else if(typeName === 'OSM') {
             d3.select('#ingestfileuploader')
             .property('multiple', 'false')
-            .attr('accept', '.osm')
+            .attr('accept', '.osm,.osm.zip')
```
brianhatchl: I think only one of Firefox and Chrome actually enforces the accept values.  But in this case, we are mandating the zipped osm file be foo.osm.zip only?  And won't support foo.zip that contains an osm file?

treycaldwell: I tested in both Firefox and Chrome (recent versions) and they will both default the file chooser to the accept types you specify.  You can still upload a .zip osm file by choosing the custom files option in the file chooser.  The file inside needs to have a .osm file extension.  I more or less did that to drive home the point this option is for zipped osm files.  Maybe that's not necessary?

Issue #466, opened 2016-05-27
MapEdit should be prevented as being selected as the secondary layer earlier


bwitham:
Currently, the check isn't performed until you click the final conflate button.  It should occur earlier.

briandaviddavidson:
I currently don't have access to this layer on my vagrant box, so I am troubleshooting with @bwitham

briandaviddavidson:
I was able to get the layer to appear :+1:

PR #569, opened 2016-07-19
466

briandaviddavidson:
An ID and class gets added to the `fieldset` tag and if the class is equal to `secondaryDataset`, underscore with remove anything with a -1 ID (in this case, the Mapedit DB layer)

I noticed at one point the ID wasn't adding correctly, but I wasn't able to recreate it. If it comes out in testing, I will work on recreating/fixing it.

Ticket: #466


sisskind:
test this please

PR #575, opened 2016-07-20
Review bubbles now draggable

jcp2236:


briandaviddavidson:
@jcp2236 Looks like this has some `eslint` errors

Try running this from the `hoot-ui` folder:
`node_modules/eslint/bin/eslint.js js/id/*` (or whatever the path to your file is)

Issue #588, opened 2016-07-25
Drawing tools stay active when dataset deleted


briandaviddavidson:
Reference Ticket: https://github.com/DigitalGlobe/VGI-team-repo/issues/177

> We broke the drawing tools, line, area... Open a dataset, try and draw a line, then close/delete the dataset from the workspace...# 1: tool stays active, #2, when you open up a new dataset and try and use line or area tools, you get this. Happens in all versions of Hoot I can currently access. Strangely it seems to be a persistent problem even after refreshing hoot and clearing cache... O_o

> ![image](https://cloud.githubusercontent.com/assets/15252833/16456259/52e40fcc-3de5-11e6-9707-991bcbe7c59e.png)

briandaviddavidson:
Able to recreate by using the steps as noted in original ticket:

1. Add a dataset
2. Click to being drawing, but instead of drawing, drag the map
3. Without stopping the drawing tool, remove the dataset

The tool stays active and trying to add another dataset results in the data never being loaded.

briandaviddavidson:
If a user is in the `draw-line` or `draw-area` mode, they will be prompted when clicking the trash can icon to finish their drawing before deleting the layer.

mikejeffe:
@briandaviddavidson I was able to trigger the error message but I can also get the "Cannot zoom.." black screen when I drag the map with an unfinished line/area draw.  I'm gonna assume that the latter condition is not a bug.

PR #589, opened 2016-07-25
588 - Prompt users to finish drawing before deleting a layer

briandaviddavidson:
If a user is in the `draw-line` or `draw-area` mode, they will be prompted when clicking the trash can icon to finish their drawing before deleting the layer.

Ticket: #588


sisskind:
@briandaviddavidson do you need to also include draw point mode?

briandaviddavidson:
@sisskind The issue doesn't happen in point mode that I've noticed. Since it's just one click "it's either there or it isn't" type of feature, it wouldn't interfere with deleting. Line and area can be half there, which is what was causing the issue.

If you want, I can add point mode as well

sisskind:
@briandaviddavidson I would like that.

PR #593, opened 2016-07-26
Measure Label Fix

sisskind:


sisskind:
@mikejeffe test on develop for measure label here (https://github.com/ngageoint/hootenanny-ui/pull/581).

PR #600, opened 2016-07-27
Text centered in review bubbles

jcp2236:


briandaviddavidson:
Text is centered in the bubbles

Issue #610, opened 2016-07-29
Modifying Review Bookmark Note Bug


sisskind:
When modifying a review bookmark, the user can click the + icon repeatedly to create multiple Cancel/Modify buttons.  Limit to one click and force user to either choose to Cancel or Modify before moving forward.  Also consider changing Modify text to Save.

![image](https://cloud.githubusercontent.com/assets/6191659/17252988/cc384092-557c-11e6-87a0-6cbe755683e5.png)

briandaviddavidson:
PR has been submitted

RobMaps:
Tested functionality of the "plus" (add) button and "cancel" button within the bookmarks window.  Upon clicking the "plus" button an avatar icon appears as well as a "cancel" and "save" button.   Upon clicking the "Cancel" button the "Cancel" and "Save" buttons do disappear however the additional avatar icon remains.  I had to hit the refresh icon to remove the additional avatar icons from the screen.  Would adding a refresh command to the "Cancel" button help?

briandaviddavidson:
Good catch, @RobMaps. I'll fix that today and submit another PR. If it gets merged this afternoon, it should be ready to test again tomorrow

briandaviddavidson:
@RobMaps The PR to fix these was merged in by @jcp2236, so you should be able to test on `develop` and not see the issue of duplicate user icons.

PR: https://github.com/ngageoint/hootenanny-ui/pull/632

RobMaps:
Sounds good.  I will retest this today.

RobMaps:
Initiated a fresh checkout of the develop branch to retest this issue.  Upon clicking the plus button 2 avatar icons appear. Hitting the cancel button only removes one of the avatar icons.  The user must refresh the screen to remove the additional avatar icons.

RobMaps:
I retested this issue today and this is now functioning properly.  Clicking the plus button only creates one avatar icon.  Clicking Cancel properly removes the additional icon(s).  Recommend that we close this issue.

RobMaps:
I retested this issue and all is functioning properly.  Clicking the plus button only adds one additional avatar icon.  Clicking the Cancel button removes the additional icon.  Recommend that we close this ticket.

PR #615, opened 2016-08-01
610

briandaviddavidson:
The user can't add multiple buttons as they could before. 

To test:

1. Create a bookmark from a merged dataset
2. Go to `Review Bookmarks`
3. Click on a bookmark
4. Click on the plus sign

Expectation: You should see the Cancel/Save buttons appear as well as an avatar icon. When clicking cancel, the avatar and buttons should go away. When clicking the plus again, they should reappear.

Ticket: #610


brianhatchl:
jenkins please run cucumber

Issue #623, opened 2016-08-02
fix help menu borders


rmrice:
"Manage Hootenanny"'s borders are off (missing bottom border, double top border, corners not rounded).

PR #625, opened 2016-08-03
623 Fixed borders in help menu

rmrice:
This is the change

Before:
screenshot from 2016-08-03 10 51 20

After:
screenshot from 2016-08-03 10 51 00

(#623)


sisskind:
lets run cucumber again

Issue #453, opened 2016-05-23
Phantom Pallet


sisskind:
Pallet goes invisible - not sure what's making it do it, both Hilary and I have seen it. You click a feature and it seems that nothing happens...except that the pallet is actually there and you can mouse over where the invisible buttons are and get the tool-tips and even click and use the pallet.

sisskind:
https://github.com/DigitalGlobe/VGI-team-repo/issues/137

sisskind:
1) Open up a dataset that has reviews
2) Go to a review that has a blue and a red feature (I have had it work with just a read feature but it wasn't reliable)
3) Click a feature so the pallet appears
4) Click and hold on another feature (I usually click a road) and then drag/pan the screen
5) Should now be experiencing a phantom pallet

briandaviddavidson:
I'm unable to recreate this on my machine, however I will continue to try throughout today. If I'm not successful, I'll reassign it to @sisskind

mikejeffe:
Also unable to reproduce this based on the steps provided.  At no point does a phantom pallet appear.

curranMapper:
I am still experiencing the phantom pallet by following the 5 steps @sisskind wrote out above

sisskind:
really? :(

sisskind:
OBE // revisit in iD 2.x

PR #626, opened 2016-08-05
#453

sisskind:
How to test (can also refer to #453)

1) Open up a dataset that has reviews
2) Go to a review that has a blue and a red feature (I have had it work with just a red feature but it wasn't reliable)
3) Click a feature so the pallet appears
4) Click and hold on another feature (I usually click a road) and then drag/pan the screen until the radial menu disappears
5) Should now ~~be experiencing a phantom pallet~~ see the radial menu again.

The fix was setting the radialMenu to undefined, which was not happening if you were clicking, holding, and panning.


brianhatchl:
I was able to reproduce this and the bug seems to occur after this js error is thrown:
```
d3.min.js:5 Uncaught TypeError: Cannot read property 'count' of undefined
```
However, there is no info in the call stack to indicate what iD or Hoot method led to this.

I'll pull the branch and see if I can still reproduce the bug or not.

sisskind:
@Brianhatchl it is in the radial_menu.js file. What is happening is the function to close the menu is being called but the svg had already been closed, hence the undefined. I thought setting ```radialmenu = undefined``` would solve this so the menu would not try to close, but apparently didn't totally cover issue.

brianhatchl:
So my above comment was just about reproducing the bug before the fix.  It is finicky to reproduce, but I have still been able to do so on the PR #453 branch.  The error is occurring after the select/drag pan, when a new feature is clicked to select it.  So would the code path be for opening the radial menu?

Again, some of the elements are actually there, just invisible because hovering over the area will reveal the tooltip for "Straighten this line" menu item.

brianhatchl:
Using the dev console inspector, the opacity property of the radial-menu <g> element is still set to zero.  I don't know if that gives a clue to where in the code something may be going wrong.

sisskind:
I think the opacity 0 is a result of the code breaking. Believe the radial menu closing script is where the error is occurring, since it is trying to close something that has already closed.

briandaviddavidson:
run cucumber

sisskind:
@brianhatchl this is due to the radial menu not properly closing (https://github.com/ngageoint/hootenanny-ui/blob/453/js/id/ui/radial_menu.js#L127)

> Using the dev console inspector, the opacity property of the radial-menu element is still set to zero. I don't know if that gives a clue to where in the code something may be going wrong.

sisskind:
@brianhatchl can you retest?  Setting ```radialMenu = undefined``` on move.select may solve problem, as it will prevent the radialMenu from trying to close in rapid succession, leading to the error.

sisskind:
test this please

brianhatchl:
If you push new commits, jenkins will retest.  No need to issue a comment.

sisskind:
Looks like cucumber test has conflict with pbf format.  Will investigate and update.

CC @briandaviddavidson @treycaldwell @brianhatchl

Issue #624, opened 2016-08-02
Review mode not showing both pairs in extent


sisskind:
Review mode should show all review features in extent.  Used to work, now it doesn't.  :(

See:
https://github.com/ngageoint/hootenanny-ui/issues/84
https://github.com/ngageoint/hootenanny-ui/issues/62

briandaviddavidson:
When going to a new pair of review features, both should now be visible on the map

curranMapper:
Went through reviews, all conflicts zoomed to extent that included both features in conflict

PR #628, opened 2016-08-05
Pan to bounds instead of entity

briandaviddavidson:
Ticket: #624 

Expectation: When going into a pair of features, both should be visible on screen.


briandaviddavidson:
jenkins please run cucumber

rmrice:
@briandaviddavidson and I tested this on my machine and it works + passes cucumber tests.

briandaviddavidson:
@brianhatchl Could this be an issue with my local vagrant box?

brianhatchl:
@briandaviddavidson I know this is late, but you should not have to rebuild your entire vagrant vm.

Next time just try restarting tomcat:
`sudo service tomcat6 restart`

brianhatchl:
I'm going to run the failing cucumber test locally.  But if it's not clear, here are the steps to prepare the PR code to be tested locally (from $HOOT_HOME on the vm host):

```
cd hoot-ui
git checkout 624
cd ..
vagrant ssh
cd hoot
source SetupEnv.sh
sudo -u tomcat6 scripts/vagrantDeployTomcat.sh
```

brianhatchl:
This is failing for me locally, on the same step, but with a different error than the jenkins job:
```
vagrant@vagrant-ubuntu-trusty-64:~/hoot/test-files/ui$ xvfb-run --server-args="-screen 0, 1024x768x24" cucumber --format progress --strict features/resolve_all_reviews.feature
/home/vagrant/.rvm/gems/ruby-2.3.0/gems/selenium-webdriver-2.53.4/lib/selenium/webdriver/common/platform.rb:166: warning: Insecure world writable dir /home/vagrant/hadoop/bin/ in PATH, mode 040777
................F--------------FF

(::) failed steps (::)

undefined method `map' for nil:NilClass
Did you mean?  tap (NoMethodError)
./features/step_definitions/custom_steps.rb:61:in `/^I should see "([^"]*)"$/'
features/resolve_all_reviews.feature:20:in `Then I should see "Complete Review"'

Failing Scenarios:
cucumber features/resolve_all_reviews.feature:3 # Scenario: I can conflate the DC GIS data and resolve all reviews at once

1 scenario (1 failed)
31 steps (1 failed, 14 skipped, 16 passed)
1m26.277s
```

brianhatchl:
I tried the cuke feature steps manually and got this error in the alert div:
```
TypeError: Cannot read property 'toString' of undefined
```
Unfortunately, the alert div seems to eat the js exception as it's not appearing in the dev console.

brianhatchl:
The review features are selected, but the map is not zooming to them due to this error.
screenshot from 2016-08-08 15 13 33

js/hoot/control/conflicts/map/FeatureHighlighter.js
```diff
@@ -151,7 +151,8 @@ Hoot.control.conflicts.map.featurehighlighter = function (context)
         _parent().reviewIds.push(relId);
         _parent().info.metadata.updateMeta(null);
         if(panToId && panTo) {
-            _parent().map.featureNavigator.panToEntity(context.entity(panToId), true);
+            var bounds = feature.loc.toString() + ',' + againstFeature.loc.toString();
```
brianhatchl: Found the cause of the cuke error.  This won't work with features that are ways.  They don't have a "loc" property.

brianhatchl:
I seem to remember working a long time ago on code that would make sure the review zoomed to the min bbox of the review features.  I'll update this with a reference to it when I can find it.

brianhatchl:
Here is the commit, https://github.com/ngageoint/hootenanny-ui/commit/bb04cbe905a9f11b4b7326fadcf8f2394bc5181a
but it appears this behavior got refactored out of existence at some point.

briandaviddavidson:
@brianhatchl Would it make sense to do separate functions to handle the different features when they go to the [`panToBounds` function](https://github.com/ngageoint/hootenanny-ui/blob/d5f91fe9931d2dbc0a13e382acb5416f82267582/js/hoot/control/conflicts/map/FeatureNavigator.js#L19) or do you think there is a better way to handle it?

brianhatchl:
It looks like the highlightLayer method specifically deals with a pair of review features, so I would just want it to call a panToBounds method with the extent of the two features.

Or a new convenience method could be added panToEntities, that accepted an array of entity ids and then calculated the extent of those and zoomed to it.

brianhatchl:
Seems like there might be a better syntax than using panToBounds.

```
map.centerZoom(extent.center(), map.extentZoom(extent));
```

You can just extend the extent of feature with review against feature and call this.  panToBounds can probably be removed entirely.

briandaviddavidson:
Wouldn't extent in this case be the entire collection of features?

briandaviddavidson:
test this please

brianhatchl:
jenkins test this please

brianhatchl:
npm test failure is caused by https://github.com/Medium/phantomjs/issues/599
will wait to see if it gets resolved upstream, but still run `npm test` locally before pushing commits please.

jcp2236:
Passed local npm test on my machine. The offset error caused by expanding the sidebar affecting the radial menu, review bubbles, etc is also affecting the centering of the review nodes here. A fix should be coming soon for that issue.

rmrice:
npm test works locally, failing on cucumber test

brianhatchl:
run cucumber

sputnam:
run cucumber

Issue #587, opened 2016-07-22
Move the Hoot UI starting location to some good looking place in the world


mattjdnv:
When a user fires up Hootenanny the initial location that displays is the DG headquarters in Westminster.

This is nice but may not work well if the user has Bing imagery turned off and does not have high resolution imagery of the westminster location available.

We should move the start location to somewhere nice that the user has a chance of seeing.

rmrice:
Any specific location requests for this? Anything anyone would like to see?

cc: @mattjdnv @briandaviddavidson @jcp2236 @sisskind @bmarchant @brianhatchl

rmrice:
Was going to do this unless we want somewhere in the US.

![screenshot from 2016-08-04 13 39 30](https://cloud.githubusercontent.com/assets/17434033/17412084/e972c422-5a48-11e6-9b16-d3b2eebf5ec8.png)

bmarchant:
NCE?  (38.753221, -77.196713) since this is ngageoint/hootenanny...  Or how about @drew-bower 's house?!

If you don't like those options, you could also randomly pick between four or five places around the globe.  I like the one you chose above, and maybe the Eiffel Tower (48.858248, 2.294566), the Great Pyramid at Giza (29.978604, 31.134543), the Statue of Liberty (40.689224, -74.044533)...

rmrice:
See PR #616

drew-bower:
@bmarchant my crib is not considered aesthetically pleasing.

mattjdnv:
@drew-bower So, you are saying that you wont be featured here http://www.mtv.com/shows/mtv-cribs ?

drew-bower:
More like the Annapolis version of this clan: https://en.m.wikipedia.org/wiki/The_Beverly_Hillbillies

PR #630, opened 2016-08-05
#587 added 7 starting locations + randomizer

rmrice:
On load you should either see:

- Palm Island, Dubai
- NCE
- Eiffel Tower
- Pyramids
- Statue of Liberty
- Sydney Opera House
- Taj Mahal


mattjdnv:
Done!

Issue #639, opened 2016-08-15
The Facility preset has the name and fcode reversed from other presets


brianhatchl:
Suggest making it consistent by editing the three AL010.json preset files for each non-OSM schema.

![screenshot from 2016-08-15 09 03 30](https://cloud.githubusercontent.com/assets/2600932/17678825/bb37294c-6306-11e6-921e-2cd8a4f71c49.png)

PR #640, opened 2016-08-15
#639 make Facility name/fcode order consistent

brianhatchl:


brianhatchl:
Boo clip.feature! jenkins re run cucumber

Issue #606, opened 2016-07-28
Sidebar expanding over map


briandaviddavidson:
This is a research ticket to try to figure out how to drag the sidebar and not move the map container to alleviate the issue of adjusting the width for all the svg elements (radial menu, drawing tools, etc.)

Related ticket: https://github.com/ngageoint/hootenanny-ui/issues/605

briandaviddavidson:
@jcp2236 is going to take over as primary since this relates to tickets he is completing

briandaviddavidson:
An initial prototype can be found on [branch `606`](https://github.com/ngageoint/hootenanny-ui/tree/606)

briandaviddavidson:
This is ready for review. The map is now 100% stationary with the bars/sidebar moving over top the map instead of moving along the same plane as the map. This solves a lot of issues involving offsetting the review bubbles, radial menu, delete menu, etc.

cc @brianhatchl @sisskind

sisskind:
Great work team!

brianhatchl:
Can you create a pull request?

curranMapper:
Yes, sidebar and top bar slide over bubbles/dial

PR #641, opened 2016-08-16
Sidebar and top/bottom bars now slide over the map

briandaviddavidson:
Before, they slid on the same plane as the map and pushed it over. Now, they slide on top of the map, like one piece of paper over another.

This should solve all the issues of having to offset all the radial menus, review bubbles, points when moving a line, etc.

Ticket: #606


brianhatchl:
@briandaviddavidson @jcp2236 Do you guys want assistance fixing this failed cuke test?

briandaviddavidson:
@brianhatchl I actually found the error and it's not with cucumber, but instead my code, so I'm in the process of fixing that now :+1:

briandaviddavidson:
Cucumber tests have passed locally.

brianhatchl:
Are you getting the jenkins email with the screenshot and build log?

briandaviddavidson:
No, I'm not. Should I be?

brianhatchl:
Yes, it should be going to the author of the pull request

brianhatchl:
testing update to jenkins job config to get culprit email working, run cucumber

brianhatchl:
another email test, run cucumber

briandaviddavidson:
Based on stepping through the cucumber test, it looks like I have to figure out another way to remove the sidebar other then using `d3.select('#someid').remove()`. Looks like that isn't preserving any of the attributes

sisskind:
Could you just hide it?

briandaviddavidson:
I would assume so. That might cause problems when it rebuilds itself though (which I believe is how it's working currently). I could look into making the whole thing a show/hide toggle.

briandaviddavidson:
Found the culprit. Looks like the necessary function (`displayHandler`) is never being called

brianhatchl:
Other places in the UI code use this:
`context.ui().sidebar.hide();`

briandaviddavidson:
Passed the original cucumber test problem, but a new one has now arisen. Going to look into that now.

briandaviddavidson:
@sisskind @brianhatchl 

The most recent failure doesn't appear to be related to my code changes. If y'all are comfortable merging my stuff in, feel free. If not, I can troubleshoot some more.

cc @jcp2236

brianhatchl:
jenkins try again to run cucumber

brianhatchl:
once more jenkins, run cucumber

briandaviddavidson:
run cucumber

brianhatchl:
This PR has broken the fix in #624.  Confirm this by running manually_resolve_reviews.feature locally.

I also noted that the centering of review features in imagery validation mode is now off by the maximum width of the sidebar, even when not expanded.
screenshot from 2016-08-25 10 46 22

briandaviddavidson:
@brianhatchl It's offset because the map had to be pushed out from behind the sidebar by 400px (default width of the sidebar). So what do you think would be the best way to fix this is? I could put the map back under the sidebar by 200px, so there would be a bit of an overlap, but the point would be centered on screen.

cc @sisskind

brianhatchl:
That makes it seem like the map is too wide then, based on where the center is being calculated.
I think I understand the intent of having the sidebar expand over top of the map, instead of causing its width to shrink.  But all the hardcoded "400"s in the diff raises a bit of code smell for me.
Because the map and all the controls within it shouldn't change position, I'd like to think the sidebar could be made expandable over top with only css changes to it and not to everything else as well.
But I don't know off the top of my head the best way to accomplish it.

sisskind:
Pure CSS may be way to go.  Hard coded values could be issue, especially with varying screen sizes.  It is also possible to attack with a callback function when sidebar is expanded to reset a context variable for adjusting radial menu, measure tool, etc for map width.

briandaviddavidson:
> But all the hardcoded "400"s in the diff raises a bit of code smell for me.

I agree, it's not the best, but the hard coded `400`s are there so the user doesn't shrink below what I thought was the best-looking minimum for the application. Anything lower than about 300 and you start getting stuff like this:

### 200px:

200px

> Pure CSS may be way to go.
Wouldn't this be considered CSS outside of the [`context.ui().sidebar.adjustMargins()`](https://github.com/ngageoint/hootenanny-ui/pull/641/files#diff-86fd9f18406cd1621c57a9eaebd4068eR79) function and related js changes being run when the sidbar is dragged (among other actions)?

brianhatchl:
Would it be acceptable if the map controls got overlapped along with the map?
All the adjustMargins code is to shrink the map controls as the sidebar expands right?
I'm not sure if the expectation is the sidebar widening would be semi-permanent for a hoot session or just to widen to view some longer field values and then shrunk back.

briandaviddavidson:
So extend the map to be the entire width of the screen and then overlay everything on top of that (in turn getting rid of the hardcoded 400px values floating around), right? That could be doable. It looks like we'd still have an issue with the offset, though. This screenshot is taking the 400px overlay off of the map (ignore the obvious blunders with the bars; I just did this in Chrome really quick):

no margin

brianhatchl:
No, I was thinking keep everything like it was before.  The map starting at the default width of the sidebar (400).  But then the map container and all it's controls stay put when the sidebar is widened on top of them.

briandaviddavidson:
Oh, I see. So instead of moving everything (like the number of reviews, the buttons across the top, etc.) it just gets sent below the sidebar until they move it back. So something like this, yes?: 

overlap

Do you think this would cause confusion if someone extends the sidebar over features (like the number of reviews left), and they don't realize it? I guess we could have the sidebar snap back after a certain period of time, but that seems like it might be too intrusive or jarring for the user. Especially if they want to keep it extended.

brianhatchl:
That's what I was thinking to avoid all the css changes that were striking me as pretty brittle.

briandaviddavidson:
I can make those changes and see if it passes cucumber/what it looks like.

briandaviddavidson:
Per suggestion from @sisskind and after talking it out, I decided to add a filler div instead of a 400px margin. The margin seems to be what was throwing it off, so everything should now center correctly and the sidebar should overlay properly.

I ran the failing test locally and it passed. I'm running all of them locally now and waiting to see what Jenkins has to say about it all.

sisskind:
I think you need to merge in latest develop (missing pbf in export options).  That should solve this issue.

briandaviddavidson:
Run cucumber

css/app.css
```diff
@@ -39,7 +39,7 @@ body {
     height: 100%;
     width: 100%;
     z-index: -10;
-    margin-left:400px;
+    /*margin-left:400px;*/
```
brianhatchl: Why was this necessary?  Removing this seems to have necessitated all the 400s below for individual child tags.

briandaviddavidson: When calculating the width of the `#content` div to 100%, it was going to the width of the entire screen. When I was adding the margin, It had already generated to be the width of the screen, so it was pushing it over and in turn, pushing 400px of the map that had already been created out of the window. That's why everything was off-center and the point wasn't being displayed on the map.

briandaviddavidson:
Going to merge this PR and start working on #676 which relies on this being added to develop

Issue #560, opened 2016-07-12
When 2 datasets are loaded, first dataset lags on refresh


sisskind:
Appears to occur with newer versions of Chrome.  Does not occur on Firefox.

Only occurs when two datasets are loaded, and it is the first dataset that was loaded that appears to lag.

sisskind:
After some further investigation, appears to only be occurring with lines or highlights, not points or polygons.

sisskind:
![560](https://cloud.githubusercontent.com/assets/6191659/17027637/a833a240-4f32-11e6-81f4-2d8ec7b22613.gif)

sisskind:
Appears to occur after the second dataset is added, but will persist even if that dataset is removed.

mikejeffe:
I wonder if this is a gpu acceleration issue:  https://bugs.chromium.org/p/chromium/issues/detail?id=117032

sisskind:
It is possible, but if that were the case I would expect it to have a lag with all features, not after adding a second layer

sisskind:
It's the ticket I just can't quit.  It is definitely layer dependent, as a created feature does not have lag until after it is saved and part of layer.  Measurement lines, unsaved features, do not lag.  The aforementioned highlight issue still exists though (but was part of previous layer, so that may work out if this bug is fixed).

![560-1](https://cloud.githubusercontent.com/assets/6191659/17036616/0ca83240-4f5b-11e6-9612-c25e8d8d1f9d.gif)

sisskind:
After testing with version prior to upstream merge, I can safely assume that this has nothing to do with the new version of iD.

sisskind:
Where paths are being returned for draw: https://github.com/ngageoint/hootenanny-ui/blob/develop/js/id/svg.js#L15

sisskind:
It is possible that it is not the feature lagging, but just the highlight.

sisskind:
It appears that it is the line-shadow that is lagging, not the line-stroke.

kweint:
Testing the lagging issue right now, and haven't found any lag on refresh when two datasets are loaded thus far. Currently testing with DC Street_Centerlines_Light and district_of_columbia_highway. Will expand testing to more datasets.

kweint:
Gave this ticket another go with every test dataset at my disposal. Could not replicate the lag bug.

PR #689, opened 2016-09-01
#560 Removing shadow on map move

sisskind:
#560 

By removing the shadow class on map move, the delayed shadow drawing no longer occurs.  This line had been commented out, but cannot find supporting evidence as to why.

### DIFF
--- a/js/id/renderer/map.js
+++ b/js/id/renderer/map.js
@@ -190,7 +190,8 @@ iD.Map = function(context) {
 
         d3.selectAll('.vertex').remove();
         //d3.selectAll('.shadow').remove();
-        d3.selectAll('path.shadow:not(.activeReviewFeature):not(.activeReviewFeature2)').remove();
+        var linesContainer = d3.select('.layer-lines');
+        linesContainer.selectAll('path.shadow:not(.activeReviewFeature):not(.activeReviewFeature2):not(.unsaved)').remove();
 
         surface
             .call(drawVertices, graph, data, filter, map.extent(), map.zoom())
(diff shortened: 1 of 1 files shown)


brianhatchl:
Can't merge this.  The review feature halos use the shadow and this commit causes the red and blue highlights to be removed.

sisskind:
@brianhatchl can you take another look?

brianhatchl:
It's not working for me.  Does it work in your testing?

brianhatchl:
The activeReviewFeature class is on the parent <g> not the <path>

sisskind:
@brianhatchl did work for me (at least I think it did).  Was able to see both features in review pair highlighted as red and blue.  Is that not what you are referring to?

brianhatchl:
yeah, I don't see halos.  they only appear when dragging the feature

sisskind:
Will give it another shot now.  Are you experiencing for all geom types or just paths?

brianhatchl:
it's POIs in my case, but the svg shadow is a path element

sisskind:
Get red/blue features, but the yellow highlight does disappear.  Fixing that.

sisskind:
So it appears that removing the shadow will break the yellow highlight for edited features.  Back to the drawing board...

sisskind:
Yes -- seeing it with POIs now too.  Really back to the drawing board this time.

sisskind:
Now see what you mean w/points where shadow is child of parent.

brianhatchl:
Looks good now.  Do we only need to remove only the line layer shadows to avoid the flickering this ticket is meant to address?

Issue #690, opened 2016-09-01
Remove WFS from UI


sisskind:
Since WFS functionality is not being used or maintained currently, remove the WFS option from the manage tab and the WFS option for export.  Once WFS is requested again, or when time allows, fix in the services tier then provide functionality through UI.

PR #695, opened 2016-09-02
Removed WFS as export option, manage tab

sisskind:
* Left remaining code intact so it will be easy to turn WFS export back on when services are fixed.
* Only code deleted was from hootManageTags.json.  This will have to be recreated when time comes.

#690 

@brianhatchl interested in your viewpoint on this -- since the code for WFS has not been deleted, it will obviously effect the code coverage stats (since tests will not touch what is not functional).  I'm inclined to not delete the code since we may turn WFS on again at some point, and it seems the least amount of risk is by keeping code in base untested.  Thoughts?


sisskind:
Will re-run cucumber test once https://github.com/ngageoint/hootenanny/pull/1027 has been merged into develop

brianhatchl:
jenkins, please run cucumber

Issue #538, opened 2016-06-29
Hoot not using full real estate


sisskind:
1. Open Chrome
2. Snap Chrome to left or right of desktop so that it fills half the desktop real estate
3. Open Hoot
4. Maximize Chrome

![image](https://cloud.githubusercontent.com/assets/6191659/16461227/83038e26-3df9-11e6-858e-fdf8b83d01f7.png)

sisskind:
https://github.com/DigitalGlobe/VGI-team-repo/issues/172

curranMapper:
boooyaaaaaa! Works, ready to close

PR #696, opened 2016-09-06
Map now re-renders when window is resized.

jcp2236:
When Hoot is opened in a minimized window, it should now scale accordingly when the window is maximized or dragged to the desired size.

Ticket: #538


sisskind:
test this please

brianhatchl:
The branch needs to merge origin/develop to fix the cucumber test failure.

PR #697, opened 2016-09-08
Fixed context menu for "zoom to layer"

rmrice:
Now "zoom to layer" context menu only appears when you right click on the title of the data layer added. 

You can test this by adding a dataset and right clicking on the layer name.

Add another dataset and conflate the two. Click on the info button and expand parameters or options, and test right clicking again. Previously, if you right clicked anywhere within the entire form containing parameters and options you would get the context menu. Now you should only get a context menu if you right click on the conflated layer title.

screenshot from 2016-09-08 16 18 05


brianhatchl:
The branch needs to merge origin/develop to fix the cucumber test failure.

PR #701, opened 2016-09-12
fix #700, copyTags is init'd to {} so fix check for it being populated

### LINKED ISSUE #700: Paste tags buttons not updating UI tag fields in sidebar
Ok, I tried it and the buttons do enable, but when clicked the tags in the sidebar are not updating.  (The change tags action does seem to be applied as the Undo button becomes enabled and there is a Change Tags step in the edit history.  I'll look into it.

brianhatchl:


brianhatchl:
jenkins run cucumber

Issue #583, opened 2016-07-21
Loading review layer from manage dataset tab should open map after pop-up


mikejeffe:
I noticed when adding a review layer that had open reviews from the Manage | Dataset list (either as reference or secondary), the pop-up appears that layer has unresolved reviews but it doesn't open the map view.  It should probably open the map since that might be confusing to the user.

sisskind:
@mikejeffe good point!  Will do for current develop branch.

RobMaps:
Completed testing for this issue.  Attempted to add a layer that still has open reviews via the data manager window.   Adding this file as the Reference or Secondary layer does not automatically change Hootenanny to the map view.  The prompt notifying the user that the layer has open reviews does automatically appear.  Further attention is needed for this issue.

sisskind:
@RobMaps @kweint Can you take a second look at this one?

kweint:
@sisskind I just tested the issue myself and had the same results.

sisskind:
@kweint thanks -- looking into now

sisskind:
@kweint can you test this one again?

kweint:
@sisskind This issue appears to be resolved!

PR #714, opened 2016-09-15
Show map when loading review layer

sisskind:
#583


brianhatchl:
failing test fixed in develop, jenkins please run cucumber

PR #715, opened 2016-09-15
376

jcp2236:
Allows edits (such as drawing lines) to be made at any zoom level without error messages. It doesn't seem to affect the drawing of any features, but maybe someone else can break it.


brianhatchl:
I noticed you need to update a mocha test for that's failing this change:
```
  1) iD.Map #zoom respects minzoom:

      expected 8 to equal 16
      + expected - actual

      -8
      +16
```

jcp2236:
Thanks, yeah looks like I'll have to update mocha and cucumber for this fix.

briandaviddavidson:
Can zoom out to level 9, then the message appears. Instead of getting stuck in the loop of not being able to zoom back in again, the user can zoom in once the message appears and go back to editing as normal. 

This works as expected.

Issue #729, opened 2016-09-15
"/" shortcut not working


curranMapper:
/ shortcut does not launch map inset

brianhatchl:
Let's add a cucumber test for this too.

briandaviddavidson:
Looks like the problem was due to the sidebar. The inset never got pushed over when the sidebar was changed to slide over the map. Should be fixed and visible now. It also slides with the map now.

kweint:
Tested on hoot NOME and current release branch so far; "/" shortcut doesn't function in either environment at the moment.

kweint:
"/" shortcut for inset doesn't function on the hoot nightly build.

kweint:
"/" shortcut works on the latest hoot nightly build!

PR #733, opened 2016-09-19
729

briandaviddavidson:
Made inset map visible as well as slide when the sidebar is expanded.

Ticket: #729


briandaviddavidson:
run cucumber

rmrice:
This works really well and works as expected. nice job @briandaviddavidson.

Issue #627, opened 2016-08-05
Context menu hangs when switching from map to manage


sisskind:
To recreate:

1) Open a context menu on Dataset tab OR tools menu in map
2) Click Manage or Return to Map
3) Context Menu remains

Context menu should disappear when no longer useful.

![image](https://cloud.githubusercontent.com/assets/6191659/17436997/0822578e-5ae9-11e6-9fe7-aecd5850e319.png)

![image](https://cloud.githubusercontent.com/assets/6191659/17436989/ffc6a360-5ae8-11e6-9541-3f2823b92f1c.png)

![image](https://cloud.githubusercontent.com/assets/6191659/17437012/1b3062ee-5ae9-11e6-900a-ed2b94380d04.png)

![image](https://cloud.githubusercontent.com/assets/6191659/17437016/2321e45a-5ae9-11e6-8c41-e3b8367f3c8e.png)

rmrice:
See also #741 which fixes context menu hang for Bulk Import
![screenshot from 2016-09-21 14 01 47](https://cloud.githubusercontent.com/assets/17434033/18722799/1493f168-8004-11e6-9cee-551b66c5ce40.png)

New PR #742

kweint:
@sisskind Could not get issue to replicate -- Context menu does not hang on hoot nightly build

PR #740, opened 2016-09-21
627: context menu hanging

rmrice:
If you return to map with the context menu still up it should now disappear :dash:

**To repeat:** 
1. Open hoot, click on manage tab
2. Right click on a dataset, any dataset!
3. Click "Return to map." Context menu should go away.

See ngageoint/hootenanny-ui#627


jcp2236:
Works great for context menu! Tools menu still sticks around when going to Manage tab.

Issue #741, opened 2016-09-21
Context menu hangs


sisskind:
Great work on #627 

Should also repeat/check for other context menus, including:

- Tools menu
- Dataset window (zoom to layer)

sisskind:
Moot point since this context menu is gone based on #759.

PR #742, opened 2016-09-21
#741: fixes context menu hang for tools menu and bulk import

rmrice:
Extension of #627. 

Fixed context menu hanging for **Tools menu**:
1. Click tools, navigate over one of the options so the second menu pops up.
2. Click "Manage" button. Context menus should disappear.

and **Bulk Import**:
1. Click on "Manage" button.
2. On datasets menu, right click "Add Dataset" to show "Bulk Import" context menu.
3. Click "Return to Map." Context menu should disappear.


jcp2236:
Nice, works as intended for me :+1:

Issue #680, opened 2016-08-31
Display Tasking Manager overlays and ME/hoot data after launching hoot iD


bwitham:
When using ME with HOT, the overlay info gets passed in as json (?) in a url from HOT.  We need to make the same thing work with hoot's iD.

This task depends on ngageoint/hootenanny-ui#679

brianhatchl:
see https://github.com/ngageoint/hootenanny-ui/commits/taskmgr

Issue #679, opened 2016-08-31
Customize Tasking Manager to launch hoot iD


bwitham:
The edit dropdown shown in HOT when you are starting your editing task should launch hoot's iD if you are working on a "conflation campaign".

ngageoint/hootenanny-ui#678, ngageoint/hootenanny-ui#684, ngageoint/hootenanny-ui#680, ngageoint/hootenanny-ui#683 are dependent on this task.

brianhatchl:
I've got the Hoot/Tasking Manager integration sorta working.

### Here are the workflow steps:

1. Ingest data to be conflated into Hoot
2. Create Tasking Manager project from menu item on that dataset
![screenshot from 2016-09-26 12 22 54](https://cloud.githubusercontent.com/assets/2600932/18843243/ec728f10-83e5-11e6-8bc9-7f1b431b3190.png)

3. Hoot POSTs complete project to tasking manager where use can view overlay of data "to be conflated" in Task Manager via node-mapnik-server
![screenshot from 2016-09-26 12 25 13](https://cloud.githubusercontent.com/assets/2600932/18843169/ad109736-83e5-11e6-9416-c89be14e2f84.png)

4. Pick a task grid from conflation project
![screenshot from 2016-09-26 12 32 10](https://cloud.githubusercontent.com/assets/2600932/18843167/aac0435a-83e5-11e6-80e3-56b79a6c97c4.png)

5. Launch Hoot that will autoload osm/me reference layer and "to be conflated" secondary layer via url hash `http://localhost:8080/hootenanny-id?#map=11/22.236328121/2.54798487714&comment=%23hootenanny-conflation-of-Congo_MGCP_Roads_Bridges&gpx=http://localhost:6543/project/3/task/11.gpx&reference=-1&secondary=161`
![screenshot from 2016-09-26 12 34 06](https://cloud.githubusercontent.com/assets/2600932/18843138/9306722a-83e5-11e6-93bd-8a0811dfa617.png)


### Next development steps:
- Need to setup osm-website/me vagrant instance with test data loaded (some place we have MGCP data to conflate, like Congo)
- Confirm that Hoot conflation will run with that data (#678)
- Do we want the launch conflation task url to configure and autostart that conflation?  Should conflation params be set as task manager project properties so users cannot adjust them?

curranMapper:
[...]

PR #750, opened 2016-09-24
Updates to integration with task manager

brianhatchl:
Adds dataset context menu item to create conflation task project
Re-enables gpx layer to display task grid #680
Add url hash handler for loading reference and secondary layers #679


brianhatchl:
See https://github.com/ngageoint/hootenanny-ui/issues/679#issuecomment-249626900

Issue #734, opened 2016-09-20
Toggle MapFeatures in iD


sisskind:
When editing in iD, and utilizing GEGD, a user may want to turn on and off several images, having the ability to turn off all features instead of clicking 12 features to better visualize the imagery. 

https://cloud.githubusercontent.com/assets/7328516/13578444/013c3100-e466-11e5-8abb-317b5e285bbc.png

briandaviddavidson:
This code is already in iD, but is being added manually due to it's high priority. We can merge it in during the next upstream merge.

PR: https://github.com/ngageoint/hootenanny-ui/pull/751

kweint:
We're solid. Toggle MapFeature is both present and functional :100:

PR #751, opened 2016-09-26
Add iD code for toggle all Map Features

briandaviddavidson:
This code is already in iD and is being added because of it's high priority

Ticket: #734


briandaviddavidson:
run cucumber

briandaviddavidson:
run cucumber

PR #752, opened 2016-09-26
Hoot-UI changes for cuc test

jcp2236:


jcp2236:
run cucumber

jcp2236:
please run cucumber

jcp2236:
jenkins run cucumber

js/id/ui/raw_tag_editor.js
```diff
@@ -186,7 +186,7 @@ iD.ui.RawTagEditor = function(context) {
             .on('change', keyChange);
 
         $items.select('input.value')
-            .value(function(d) { return d.value; })
+            .attr('value', function(d) { return d.value; })
```
brianhatchl: What is this change fixing?  These lines seems like they would be equivalent.

brianhatchl:
Also, to fix the failing cuke test, merge origin/develop into this branch.

js/id/ui/raw_tag_editor.js
jcp2236: It's similar, but adding the value as an attribute allowed cucumber to identify the element in a way that wasn't working using .value. It shows up as an attribute for the element as seen in this screen grab:
screenshot from 2016-09-27 10 37 59

jcp2236:
I must have merged before the fixes that came in yesterday. Will merge again.

js/id/ui/raw_tag_editor.js
brianhatchl: Okay, we need to make sure that is equivalent then.  Seems like that change might run a risk of breaking the functionality of the input.  Would be better to write a cucumber custom step that checks the value even if it can't be expressed in a simple css selector.

jcp2236: If I kept both lines of code (and changed the name of the attr), would it be too bloated? I couldn't get cucumber to check the value unless it was a css selector.

brianhatchl: Try this step, Then I should see element "cssSelector" with value "value"

jcp2236: I couldn't get that method to work before, but I rewrote my cuc test, and it does now. Not sure what I missed before. I'll close this PR and open up one for the cuc test, no changes to any js files needed.

jcp2236:
No changes were made, closing PR.

Issue #694, opened 2016-09-02
Do not allow conflation job if outstanding edits exist


sisskind:
_I believe this should be the case, but please sound the alarms if I am wrong._

Hootenanny should prevent a conflation job from occurring within the UI if there are outstanding edits that have not been saved.  This is similar to preventing a user from moving on from a review if edits exist while in review mode.

CC @brianhatchl @mattjdnv @bwitham

brianhatchl:
Sounds like correct desired behavior to me.

PR #777, opened 2016-10-19
#694 check for unsaved edits before conflating

rmrice:
#694 

To make sure this works:
1. add 2 datasets
2. make some edits
3. do not save them
4. press "conflate"

you should see an error alert pop up that warns you to save your edits before conflating.

Then you can either save or undo the edits and conflation should work.

Thanks @sisskind for your assistance :smile:


sisskind:
Using window.id.  Not happy about it, but seems to be best implementation right now.  Issue is the closure function not accessing global variables in certain places.  May require longer look at entire code base.

PR #784, opened 2016-11-07
756

jcp2236:
Changed to arrows

Issue #756, opened 2016-09-27
Change 'Previous' and 'Next' to r/l arrows


jcp2236:
Until a fix for the behavior of progressing through review items is made, they'll be changed to arrows.

kweint:
Tested on hoot nightly -- still shortcuts are still p/n. Repeated test in NOME hoot, for giggles, with same results.

sisskind:
OBE

Issue #304, opened 2016-03-09
Blue attribute box mostly off screen in some instances


curranMapper:
note from 3/9 training session:
Was not intuitive that when the blue attribute box appeared mostly off screen during conflict resolution was because there were no associated tags.  Is there a clearer way of visualizing so user doesn't think something is "wrong" in these instances?

jcp2236:
CSS fixes to widen blue attribute boxes/column in #798

kweint:
Tested on hoot nightly and the blue attribute box appears to look a-ok.

PR #792, opened 2016-11-17
304

jcp2236:
Add placeholder text "N/A" in review table for conflated features

### DIFF
--- a/dist/img/iD-sprite.svg
+++ b/dist/img/iD-sprite.svg
@@ -9,7 +9,7 @@
       </cc:Work>
     </rdf:RDF>
   </metadata>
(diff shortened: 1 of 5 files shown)


brianhatchl:
I'm not sure I like the 'N/A' solution to #304.
screenshot from 2016-11-23 11 20 32
These could easily be confused with tag values IMO.

Maybe non-breaking spaces `&nbsp;` would make the column wider but still appear blank instead?

Also, this PR branch appears to have commits from https://github.com/ngageoint/hootenanny-ui/pull/784 mixed in, which should be removed.

jcp2236:
Good idea, I like the idea of widening the column. I'll get the old commits removed, I noticed that the other day too. I haven't deleted old commits before, should I use the command `git rebase -i HEAD~4` or am I way off track?

brianhatchl:
When working on a new feature, you should create a new feature branch off of the develop branch before making any changes and commits:
```
git checkout develop
git pull
git checkout -b my_new_feature_branch
```
Maybe you did a new branch checkout from the other feature branch instead of develop?

jcp2236:
Yeah, I was working on both, and it looks like I got the 'wires crossed'.

jcp2236:
I was asking about how to delete the two earliest commits in this branch, which were meant for #756, not #304

brianhatchl:
might just be best to recreate the branch and cherry-pick the intended commit.  either way git history would get changed, so deleting the remote branch will be necessary too.  not sure if the pull request needs to be deleted as well when removing the remote branch.

jcp2236:
Alright, I thought that might be the way I had to go. Thanks for the help.

Issue #758, opened 2016-10-02
Lookup FCODE/Name on Import


sisskind:
On import, provide the ability for **known** translations to look up FCODE and append english name.  For example, if importing LP010, name dataset LP010_Road.

FYI @mattjdnv @brianhatchl

kweint:
Cool news: This neat new feature appears to be working!

PR #795, opened 2016-11-18
Lookup FCODE/Name on Import

sisskind:
refs #758


brianhatchl:
jenkins run npm test

sisskind:
found npm error.  will fix and recommit.

brianhatchl:
jenkins run cucumber

brianhatchl:
jenkins run npm test

js/hoot/control/utilities/ImportDirectory.js
```diff
@@ -317,7 +371,12 @@ Hoot.control.utilities.importdirectory = function(context) {
     var _highlightOption = function(optName,status) {
         var selectedOpt = d3.select('#importDirectoryFilesList').select('option[value="' + optName + '"]');
 
+        if(_.isEmpty(selectedOpt[0][0])){
```
brianhatchl: There is a d3 method selection.empty() that might work better here

js/hoot/control/utilities/ImportDirectory.js
```diff
@@ -880,6 +946,38 @@ Hoot.control.utilities.importdirectory = function(context) {
...
+                selOpt.text(fcodeName);
+            }
+        },fcodeList);
+        _validateFileList(_getFilesList());
+    };
+
+    var removeFCodeDescription = function() {
+        // Replace text of each option with value (original value)
+        var opts = d3.select('#importDirectoryFilesList').selectAll('option')[0];
+        _.each(opts,function(opt){
```
brianhatchl: I'm bothered by the use of [0] here and then lodash and other array methods when I think there is probably a d3 selection method that will do the same thing.  It feels hacky.

js/hoot/Hoot.js
```diff
@@ -301,6 +301,12 @@ Hoot.hoot = function (context) {
         } catch (err) { return false; }
     };
 
+    hoot.removeSpecialChar = function(str){
+        var pattern = new RegExp(/[~`!#$%\^&*+=\-\[\]\\';,/{}|\\":<>\?|]/); 
+        var retval = str.replace(pattern,'');
+        return retval;
+    };
+
```
brianhatchl: Let's be sure to write a mocha test for this.

js/hoot/control/utilities/ImportDirectory.js
```diff
@@ -168,10 +179,52 @@ Hoot.control.utilities.importdirectory = function(context) {
...
+        return retval;
+    };
+
+    var _getDescriptionList = function(translation){
+        if(!translation){translation = _getTranslation();}
+
+        d3.xhr(window.location.protocol + '//' + window.location.hostname +
+            Hoot.model.REST.formatNodeJsPortOrPath(iD.data.hootConfig.translationServerPort)
+        +'/schema?translation='+translation+'&limit=10000')
+        .get(function(error, resp){
```
brianhatchl: We don't really need to make a call to the translation server when we're removing the fcode descriptions.  This class should also probably cache the response to only have to make the call once.

js/hoot/control/utilities/ImportDirectory.js
```diff
@@ -168,10 +179,52 @@ Hoot.control.utilities.importdirectory = function(context) {
...
+        var filesList = _.map(d3.select('#importDirectoryFilesList').selectAll('option')[0],function(opt){return {value:opt.value,text:opt.text};});
         return filesList;
     };
 
+    var _checkForDescription = function(){
+        var mgcpCheck = d3.select('#importDirectorySchema').value().indexOf('MGCP');
+        var tdsCheck = d3.select('#importDirectorySchema').value().indexOf('TDS');
+        var cboxBool = mgcpCheck > -1 || tdsCheck > -1;
+        d3.select('.cboxAppendFCode').classed('hidden',!cboxBool).select('input').property('checked',false);
+        _getDescriptionList('MGCP');
```
brianhatchl: This arg sometimes needs to be TDS.

I'd almost rather this method set an internal variable that can be used instead of the code in _getTranslation()

js/hoot/control/utilities/ImportDirectory.js
```diff
@@ -168,10 +179,52 @@ Hoot.control.utilities.importdirectory = function(context) {
...
+
+        return retval;
+    };
+
+    var _getDescriptionList = function(translation){
+        if(!translation){translation = _getTranslation();}
+
+        d3.xhr(window.location.protocol + '//' + window.location.hostname +
+            Hoot.model.REST.formatNodeJsPortOrPath(iD.data.hootConfig.translationServerPort)
+        +'/schema?translation='+translation+'&limit=10000')
```
brianhatchl: I think we can remove the limit param since we've fixed the default in the server to be 1000.

js/hoot/control/utilities/ImportDirectory.js
```diff
@@ -880,6 +946,38 @@ Hoot.control.utilities.importdirectory = function(context) {
...
 
+    /**
+    * @desc Function to append FCode Description based on FCODE
+    * @desc Assumes that layer name is a valid FCODE
+    * @oaram fcodeList - JSON of FCODEs and Descriptions
+    * @param filesList - Selected files list.
+    **/
+    var appendFCodeDescription = function(fcodeList) {
+        var filesList = _getFilesList();
+        _.each(filesList, function(f){
```
brianhatchl: I think you could do this more cleanly with a d3 selection of the option elements.

js/hoot/Hoot.js
sisskind: https://github.com/ngageoint/hootenanny-ui/issues/796

js/hoot/control/utilities/ImportDirectory.js
sisskind: Will use Internal. Had set it there to return something in try catch as placeholder and forgot to change before committing code. Will return nothing with error message. Thanks for the catch.

js/hoot/control/utilities/ImportDirectory.js
sisskind: Thanks - will give that a shot.

js/hoot/control/utilities/ImportDirectory.js
sisskind: think it might be best to add a class to the options and then run a selectAll on that class. Will implement this change.

js/hoot/control/utilities/ImportDirectory.js
sisskind: That's because it is hacky. Will reevaluate.

sisskind:
run cucumber

sisskind:
Closing this PR in preference for #800 which includes this code base.

PR #797, opened 2016-11-28
use proxy host for egd to avoid basic auth issues

brianhatchl:
also add switch support to wfs request (if dg services ever
support CORS)


brianhatchl:
@rmrice can you review and merge these changes for me?

PR #798, opened 2016-11-29
Remove JS changes for tag table, adjust size of columns

jcp2236:
So I widened the columns, they default to the width in the image below (75px) when there's no data in the tag table. Let me know if this looks better than "N/A" as a placeholder.

screenshot from 2016-11-29 11 03 42


jcp2236:
run cucumber

PR #804, opened 2016-12-13
708

jcp2236:
Updated translation function for copying tags in iD


jcp2236:
This branch has some crossover

Issue #787, opened 2016-11-09
Can't save changes after adding 2 datasets


sisskind:
Hoot doesn't like it when you draw new data (like a new point feature) after adding 2 datasets:

![](https://cloud.githubusercontent.com/assets/17434033/20067527/86c4d276-a4e3-11e6-865c-6e564252dddc.png)

It perpetually is "uploading changes to hootenanny"

Had @jcp2236 test this out and he's having the same problem. Saving changes works fine if you only have one dataset on the map, but if you have two datasets on the map it freaks out. I think the problem is it doesn't know which dataset to add the changes to. Potential solutions could be to provide the capability of designating which dataset the newly drawn data feature gets added to, or adding the newly drawn data as a third dataset.

@sisskind is this a legitimate bug?

rmrice:
Taking a look at this now. I can't find a warning if there was one, so I'm thinking that if two datasets are visible on the map, gray out the point/line/area and add a tooltip asking to turn off one of the layers.

sisskind:
Change so that you can edit with 2 datasets.

brianhatchl:
My recollection of how this used to behave is that when attempting to add a new feature (and maybe edit an existing feature as well) the user was presented with a popup that asked them to uncheck the other layer so that only the edited layer was visible.  Then editing would work, though I'm not sure where the code was that identified which layer was visible and thus editable, that may be gone too.

So not sure if this is the desired behavior for restoring editing functionality with two layers loaded, but at least it's some baseline information.

kweint:
Provided I 'turn off' one of the layers to designate the active layer I want the edits saved to, I'm not encountering any errors.

PR #811, opened 2016-12-27
Error message resurrected when trying to save with 2 layers present

sisskind:
* Previous code checked newWays length with vis.length to change go, but not sure why.  Changed to throw error if vis.length !== 1.

#787


sisskind:
run cucumber

sputnam:
run cucumber

Issue #824, opened 2017-01-11
Hiding map features doesn't work for conflated layer in review mode


brianhatchl:

sisskind:
Is this after conflation, when layer is reloaded, or both?

brianhatchl:
Both.

brianhatchl:
But it does work on merged layers with reviews when not in review mode.

sisskind:
So it sounds more like a review mode issue? Can you hide features from input layers if they are turned on?

brianhatchl:
Nope.

brianhatchl:
When you past self sets a booby trap for your future self:

From Oct. 2015 https://github.com/ngageoint/hootenanny-ui/commit/7a12130

(internal) https://10.194.71.155/redmine/issues/7155
### POIs under review not consistently displaying at default zoom level

>So this issue impacts drawVector operations, not just drawVectorFar. It seems like iD has a native filtering that will autohide certain features at certain scales. We got around this with loadMissing when the tiled osm data had too many nodes, but another problem reveals itself when the dataset is big enough to load all at once, which is the case for the sample data in this ticket. In that case the data is already loaded, so loadMissing is not invoked, but the point features are autoHidden. The code doing this seemed pretty tangled to me, so the simplest approach was to just disable this filtering when in review mode (the same way we disable drawVectorFar).

Will have to see if we can disable autohidden without disabling the ability of the Map Data / Map Features control to hide selected features.

kweint:
Tested on Hoot nightly version and was not able to replicate issue. Deeming this a success!

PR #826, opened 2017-01-12
allow map feature filtering in review mode

brianhatchl:
add review features (points) back in if they get filtered out

refs #824


brianhatchl:
In order to get the Map Features filtering to work, we could not skip the features.filter() call, but we could add the review features back within that filter method.

PR #834, opened 2017-01-26
Updated hoot logo

rmrice:
screenshot from 2017-01-26 09 57 28


rmrice:
Doesn't work when deployed on hoot-server. Investigating.

js/hoot/view/utilities/Utilities.js
```diff
@@ -173,34 +173,15 @@ Hoot.view.utilities = function (context){
...
+        .classed('point', true)
         .attr('href', '#version')
         .append('img')
-        .classed('col1 cursor', true)
-        .attr('src', 'img/hoot_logo_icon.png')
-        .attr('height', '60px');
-        labelContainer.append('div')
-        .classed('inline dark strong hoot_label cursor col2', true)
-        .html('Hootenanny')
+        .attr('src', context.asset('dist/img/logo/hoot_logo_update.png'))
```
brianhatchl: I don't think you want the 'dist' in the path here.  That's what context.asset( ) adds if necessary.

Does this work when deployed via hootserver and tomcat?

rmrice: No, the I discovered that the img src has to be defined in the css. It should work with the latest commits.

rmrice:
This should work now.

PR #865, opened 2017-02-21
Update header bar

sisskind:


modules/ui/sidebar.js
```diff
@@ -11,6 +11,16 @@ export function uiSidebar(context) {
 
 
     function sidebar(selection) {
+        var labelContainer = selection.append('div').classed('fill-dark',true);
+
+        labelContainer.append('div')
+            .attr('href', '#version')
+            .classed('point hoot_label', true)
+            .attr('height', '60px')
```
brianhatchl: move this prop to css I think

img
```diff
@@ -0,0 +1 @@
+dist/img
```
brianhatchl: build.js should create this symlink, so remove it

css/img
```diff
@@ -0,0 +1 @@
+../dist/img
```
brianhatchl: build.js should create this symlink, so remove it

sisskind:
@brianhatchl we need to look at the build.js then, because I'm not sure it is 100% doing what we expect

PR #868, opened 2017-03-01
Fix Advanced Options Label

rmrice:
Label stays where it's supposed to now :see_no_evil:

### DIFF
--- a/js/hoot/control/Conflate.js
+++ b/js/hoot/control/Conflate.js
@@ -618,7 +618,7 @@ Hoot.control.conflate = function (context,sidebar) {
         ];
 
     };
-    
+
     /**
     * @desc Help function for new name generation
     * @param words -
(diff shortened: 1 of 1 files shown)


rmrice:
run cucumber

rmrice:
```
error: Your local changes to the following files would be overwritten by checkout:
	missing
Please, commit your changes or stash them before you can switch branches.
Aborting
```

???

brianhatchl:
run cucumber

brianhatchl:
@rmrice Just like the error states, the git repo on the jenkins vm had a modifed file tracked by git `missing` (which is used by the c++ compiler I think).  The jenkins job does a `git pull` from the develop branch and it was unable to do so because the `missing` file was modified and would be overwritten.  Same thing would happen on your local git repo if you had changed tracked files and tried to pull.

I solved it this once by ssh'ing to the jenkins vm and doing `git checkout -- missing` to restore the state of the file from the most recent commit so that git pull would succeed.

PR #874, opened 2017-04-17
Update alert div to handle formatting for UI errors, read in error output from back-end

jcp2236:
Errors/progress/success messages should display with these changes, but including further details about Hoot core commands has been pushed to another ticket inside a new UI error handling Epic.

### DIFF
--- a/js/hoot/model/rest.js
+++ b/js/hoot/model/rest.js
@@ -398,7 +398,6 @@ Hoot.model.REST = function (command, data, callback, option) {
                 if (resp.status === 'complete') {
                     var prettyString = function(obj) {
                         var finalString = '';
-                        console.log(obj);
                         for (var atr in obj) {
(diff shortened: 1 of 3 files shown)


js/hoot/model/rest.js
```diff
@@ -393,6 +394,25 @@ Hoot.model.REST = function (command, data, callback, option) {
 
                     JobStatusStopTimer(resp);
                 }
+
+                if (resp.status === 'complete') {
+                    var prettyString = function(obj) {
+                        var finalString = '';
+                        for (var atr in obj) {
+                            // if (atr === 'commandDetail') {
```
sisskind: Any reason to keep commented out section?

jcp2236: You're right, that's likely going in the modal window now, or a similar function. I'll take it out.

PR #877, opened 2017-04-26
Add in the core command detail modal window and Hoot-consistent styling

jcp2236:


js/id/ui/alert.js
```diff
@@ -1,12 +1,14 @@
 iD.ui.Alert = function(message,type,stack) {
+    var modal = document.getElementById('myModal');
+    // var btn = document.getElementById("myBtn");
```
briandaviddavidson: Let's remove these variables if they aren't being used

js/id/ui/alert.js
```diff
@@ -20,12 +22,31 @@ iD.ui.Alert = function(message,type,stack) {
...
+            .attr('id','detailDiv')
+            .classed('fill-darken3 pin-top pin-left pin-bottom pin-right', true);
+        var detailModal = modalbg.append('div')
+            .classed('contain col8 pad1 hoot-menu fill-white round detailModal', true);
+        detailModal.append('h1')
+            .attr('id','processingDivLabel')
+            .style({'text-align': 'center'})
+            .text('Hoot Core Command Details');
+    }
+    window.onclick = function(event) {
```
briandaviddavidson: Should we do this as a window click or a click handler on `#detailDiv`? Might make more sense to do it on `#detailDiv` since 1. it takes up the whole screen and 2. we don't have to worry about the `window.onclick` interfering with anything else in the browser.

cc @sisskind

jcp2236: Yeah, you're right. It's consistent with the rest of Hoot's code to make it a d3-based click handler.

PR #885, opened 2017-05-03
873

jcp2236:


jcp2236:
jenkins run cucumber

js/id/ui/alert.js
```diff
@@ -21,29 +20,67 @@ iD.ui.Alert = function(message,type,stack) {
     if(type === 'success'){alertDiv.classed('green',true);}
 
     function modalWindow() {
+        d3.select('#detailDiv').remove();
         var modalbg = d3.select('body').append('div')
             .attr('id','detailDiv')
             .classed('fill-darken3 pin-top pin-left pin-bottom pin-right', true)
-            .on('click', function() {
+            .on('click',function() {
                 if (d3.event.target === d3.select('#detailDiv').node()) {
```
brianhatchl: Do we really need this if statement?  Wouldn't `d3.select(this).remove()` always work here?

js/id/ui/alert.js
```diff
@@ -21,29 +20,67 @@ iD.ui.Alert = function(message,type,stack) {
...
             .text('Hoot Core Command Details');
+        detailModal.append('p')
+            .html(fullDetail);
     }
 
+    var prettyString = function(obj) {
+        var finalString = '';
+        for (var atr in obj) {
+            if (atr === 'commandDetail') {
+                obj.commandDetail =  JSON.stringify(obj.commandDetail, null, '<br>');
```
brianhatchl: Couldn't this line be set without the if statement outside the for loop?

js/id/ui/alert.js
```diff
@@ -21,29 +20,67 @@ iD.ui.Alert = function(message,type,stack) {
...
+        return '<br>' + finalString;
+    };
+
+    var fullDetail = prettyString(message);
+
+    var removeDetail = function(obj) {
+        var lessDetail = _.clone(obj);
+        for (var param in lessDetail) {
+            if (param === 'commandDetail') {
+                delete lessDetail.commandDetail;
```
brianhatchl: Couldn't the delete statement be done without the for loop and if statement?

js/id/ui/alert.js
jcp2236: Pretty sure I added the if statement because the window was closing when you clicked both on it and outside of it, instead of just outside of it, as intended. I'll remove it and test, you may be right.

js/id/ui/alert.js
jcp2236: Yep, I'll refactor.

js/id/ui/alert.js
jcp2236: Just tested, if it's taken out clicking anywhere removes the div instead of just the 'x' and outside the window.

jcp2236:
jenkins run npm-test

jcp2236:
jenkins run npm test

js/id/ui/alert.js
```diff
@@ -44,10 +44,8 @@ iD.ui.Alert = function(message,type,stack) {
 
     var prettyString = function(obj) {
         var finalString = '';
+        obj.commandDetail =  JSON.stringify(obj.commandDetail, null, '<br>');
```
brianhatchl: This change seems to be the culprit that breaks tests.  Not sure why, but I guess put the line back in the for loop.

jcp2236: Weird. Thanks for taking a look, I'll put that back inside the loop.

jcp2236: So without that if statement, it loops through and parses the JSON multiple times, making for a really ugly output (see below). The original code was how I got around that. If it weren't for the failing mocha tests, your simpler solution makes for better logic, not sure what the deal is.
screenshot from 2017-05-04 14 59 04

brianhatchl: I meant just to revert the recommended change, so both for loop and if stmt like it was before.

jcp2236:
jenkins run cucumber

brianhatchl:
@jcp2236 Got the same cucumber failure twice, so it may legit.  You should run the test locally and see if it can be reproduced in isolation and if so is it a regression bug or a change in behavior that requires the test to be updated.

jcp2236:
That specific test passed locally for me. I thought my changes may have made an impact, but I saw the new error message I created and the test passed no problem.

PR #893, opened 2017-06-05
hoot:status may be undefined

brianhatchl:
this may be the true cause of
https://github.com/ngageoint/hootenanny/issues/1602

the intermittent failure probably depends on what review is loaded first


brianhatchl:
run cucumber

brianhatchl:
run cucumber

brianhatchl:
run cucumber

Issue #927, opened 2017-08-22
Later versions of make remove sort order relied on by dist/iD.js


brianhatchl:
This is the symptom of incorrect js bundling order in dist/iD.js:
>iD.js:30109 Uncaught TypeError: Cannot set property 'relation' of undefined
   at iD.js:30109
   at iD.js:47640
(anonymous) @ iD.js:30109
(anonymous) @ iD.js:47640
presets.js:1 Uncaught TypeError: Cannot set property 'presets' of undefined
   at presets.js:1
(anonymous) @ presets.js:1
imagery.js:1 Uncaught TypeError: Cannot set property 'imagery' of undefined
   at imagery.js:1
(anonymous) @ imagery.js:1
(index):27 Uncaught TypeError: Cannot read property 'buildInfo' of undefined
   at (index):27

Make removed sort order of filename wildcard expansion:
>The NEWS file for make-3.82 says:
>* WARNING: Future backward-incompatibility!
 Wildcards are not documented as returning sorted values, but up to and
 including this release the results have been sorted and some makefiles are
 apparently depending on that.  In the next release of GNU make, for
 performance reasons, we may remove that sorting.  If your makefiles
 require sorted results from wildcard expansions, use the $(sort ...)
 function to request it explicitly.

Unbuntu 14.04 uses make 3.81, so it works.
But Ubuntu 16.04 uses make 4.1 and CentOs 7 uses make 3.82, so dist/iD.js is broken when built in those environments.

Removing wildcards from the js source files for dist/iD.js will fix this.

PR #928, opened 2017-08-22
remove source file wildcards

brianhatchl:
refs #927


brianhatchl:
@mattjdnv went ahead and merged, there is an expected cuke failure we are tracking in another ticket.

mattjdnv:
OK. Sounds good.

Issue #944, opened 2017-11-21
Include last accessed date in data table & set warnings for old datasets


rmrice:
Goes with https://github.com/ngageoint/hootenanny/issues/1979

rmrice:
to do:

- [x] add last accessed date to data table
- [x] add moment.js and use it to have last accessed date display as time ago
- [ ] create a method that selects all datasets that are last accessed once they hit 3 months, and bulk delete those (up for discussion)

rmrice:
So this is almost done, I am just working on adding a tooltip to the svgs (little alarm clocks next to datasets older than 2 months).

When the datasettable is built, for each node that has a last accessed date of 3 months, the data will pass through `context.hoot().model.layers.deleteLayer()` to get deleted. I think this is pretty straight forward and seems to work. It is difficult to test since I am at the mercy of local data I have added to my hoot box.

rmrice:
Going to hold off on auto-deleting old datasets until this is discussed more.

We have a few options:
- Create an inconspicuous UI element (button) that allows a authoritative user to bulk delete all datasets
- Improve UI of the dataset table so that a user can bulk select datasets for deletion and sort by last accessed

PR #951, opened 2017-12-04
Let users know if dataset is >=2 months old; add header to dataset table

rmrice:
Refs 
https://github.com/ngageoint/hootenanny-ui/issues/944
https://github.com/DigitalGlobe/VGI-team-repo/issues/281
https://github.com/ngageoint/hootenanny/issues/1979

Next step will be refactoring table to improve UX for selecting older datasets for deletion.


build.js
```diff
@@ -3,6 +3,7 @@ var fs = require('fs'),
     glob = require('glob'),
     YAML = require('js-yaml'),
     _ = require('./js/lib/lodash'),
+    moment = require('./js/lib/moment'),
```
brianhatchl: I don't think we need this, since we use the js/lib file

Makefile
```diff
@@ -440,4 +441,4 @@ js/lib/d3.v3.js: $(D3_FILES)
 	@echo 'd3 rebuilt. Please reapply 7e2485d, 4da529f, and 223974d'
 
 js/lib/lodash.js: Makefile
-	node_modules/.bin/lodash --development --output $@ include="any,assign,bind,chunk,clone,compact,contains,debounce,difference,each,every,extend,filter,find,first,forEach,forOwn,groupBy,indexOf,intersection,isEmpty,isEqual,isFunction,keys,last,map,omit,pairs,pluck,reject,some,throttle,union,uniq,unique,values,without,flatten,value,chain,cloneDeep,merge,pick,reduce" exports="global,node"
+	node_modules/.bin/lodash --development --output $@ include="any,assign,bind,chunk,clone,compact,contains,debounce,difference,each,every,extend,filter,find,first,forEach,forOwn,groupBy,indexOf,intersection,isEmpty,isEqual,isFunction,keys,last,map,omit,pairs,pluck,reject,some,throttle,union,uniq,unique,values,without,flatten,value,chain,cloneDeep,merge,pick,reduce" exports="global,node"
```
brianhatchl: let's add the newline back in.

js/hoot/control/utilities/Folder.js
```diff
@@ -139,6 +139,32 @@ Hoot.control.utilities.folder = function(context) {
...
 
+        function updateLastAccessed(node) {
+            var row = d3.select(node.parentNode);
+            row.classed('expiring', true)
+                .append('g').append('svg:foreignObject')
+                .attr('class', 'expiring')
+                .attr('transform', function(d) {
+                    var dd = d.depth-1;
+                    var dy=772+(11*dd);
+                    return 'translate('+ dy +',-9)';
```
brianhatchl: The timer positioning gets janky if we resize the window.
image

js/hoot/control/utilities/Folder.js
```diff
@@ -139,6 +139,32 @@ Hoot.control.utilities.folder = function(context) {
...
+            row.classed('expiring', true)
+                .append('g').append('svg:foreignObject')
+                .attr('class', 'expiring')
+                .attr('transform', function(d) {
+                    var dd = d.depth-1;
+                    var dy=772+(11*dd);
+                    return 'translate('+ dy +',-9)'; 
+                });
+            var tooltip = d3.select('#datasettable')
+                .append('div')
```
brianhatchl: Seems like the tooltip div should only get appended once.  This will append a new div every time the method is called.  You can fix that by adding a data binding of `.data([0]).enter()` so that there will only be one tooltip.

js/hoot/control/utilities/Folder.js
```diff
@@ -139,6 +139,32 @@ Hoot.control.utilities.folder = function(context) {
...
+                .attr('class', 'tooltip-old-dataset')
+                .text('This dataset has not been used in a while.');
+            d3.selectAll('g.node.expiring')
```
brianhatchl: could you add this on 145 above right after you `.append('g')`

js/hoot/control/utilities/Folder.js
```diff
@@ -139,6 +139,32 @@ Hoot.control.utilities.folder = function(context) {
...
+                .on('mousemove', function(){
+                    tooltip
+                        .style('left', Math.max(0, d3.event.pageX - 200) + 'px')
+                        .style('top', (d3.event.pageY - 50) + 'px')
+                        .style('opacity', '0.9');
+                })
+                .on('mouseout', function () {
+                    return tooltip.style('opacity', 0);
```
brianhatchl: not sure you need to `return` anything here.

package.json
```diff
@@ -52,5 +52,8 @@
   "engines": {
     "node": ">=0.10.0"
   },
-  "engineStrict": true
+  "engineStrict": true,
+  "dependencies": {
+    "moment": "^2.19.2"
+  }
```
brianhatchl: Do we need this dependency when we are using the js/lib file?

js/hoot/view/utilities/Dataset.js
```diff
@@ -73,10 +73,19 @@ Hoot.view.utilities.dataset = function(context)
...
-        form.append('div')
+        var table = form.append('div')
         .attr('id','datasettable')
-            .classed('col12 fill-white small strong row10 overflow keyline-all', true)
-            .call(hoot_view_utilities_dataset.populateDatasetsSVG);
+            .classed('col12 fill-white small strong row10 overflow keyline-all', true);
+
+        var datasetColHeaders = [{title: 'Dataset', class: '14%'}, {title: 'Last Accessed', class: '7%'}, {title: 'Import Date', class: '6%'}, {title: 'Size', class: '1%'}];
+        table.insert('div').attr('id', 'datasettableHeader')
+            .append('span').selectAll('th')
```
brianhatchl: Do we need the span when we already have a div container?

js/hoot/control/utilities/Folder.js
```diff
@@ -204,18 +230,31 @@ Hoot.control.utilities.folder = function(context) {
...
+                      .style('fill', fontColor)
+                      .attr('dy', 3.5)
+                      .attr('dx', function() { return '50%'; })
+                      .attr('text.anchor', 'end')
+                      .text(function(d) {
+                            var lastAccessed = d.lastAccessed,
+                                timeAgo = lastAccessed.replace(/[-:]/g, '');
+                            var dateActive = moment(timeAgo).fromNow();
+                            // show warning if dataset hasn't been accessed in 2 months
+                            var oldData = dateActive.split(' ')[1] === 'months' && dateActive.split(' ')[0] >= 2;
```
brianhatchl: This isn't very robust.  What is there is data last accessed "a year ago".  Let's not use the fromNow() value for this conditional, let's use something like `moment.now().diff(moment(timeAgo), 'days') > 60`

brianhatchl:
@rmrice p.s. you were right that I could not get the title attribute or tag to work here.

package.json
rmrice: I don't think so, but might be smart to keep in there for when we move to hoot 2.0.

js/hoot/control/utilities/Folder.js
rmrice: I was building this under the assumption we wouldn't have any datasets that were >1 year old (based on introducing lastAccessed and auto-deleting datasets that are 3 months or older). Now that we are not auto deleting though, you are right.

js/hoot/control/utilities/Folder.js
rmrice: line 145 is appending the alarm clock icon. The tooltip behavior is getting added to the row classed `g.node.expiring`

js/hoot/control/utilities/Folder.js
rmrice: I couldn't get this method to work.

js/hoot/view/utilities/Dataset.js
rmrice: I guess not.

PR #968, opened 2018-02-06
Add section for differential conflation to stats panel

mschicker:
refs https://github.com/ngageoint/hootenanny/issues/2084


mschicker:
@brianhatchl - Could you glance over this when you get a moment?

PR #1196, opened 2018-08-20
develop

mschicker:


mschicker:
This is not what I wanted at all. The wrong button had focus when I hit "enter" :(

PR #1197, opened 2018-08-20
Add differential conflation w/Tags to the UI

mschicker:
Refs hoot ticket #2525


mschicker:
I see. I <3 submodules.

PR #1238, opened 2018-10-04
bug fix: oauth: handle authentication failure gracefully

sumarlidason:
refs #1237

* add padding to bottom of login page
* pass back errors from popup window console to main window console for easier debugging
* fix issue w/ infinite refresh loop
* remove oauth params from address bar when they fail

### DIFF
--- a/css/login.css
+++ b/css/login.css
@@ -23,7 +23,7 @@ div.hoot-login-cell.text {
 }
 div.hoot-login-container {
     margin: 0px 15% 0px 15%;
-    padding: 10px;
+    padding: 10px 10px 60px 10px;
     width: 70%;
 }
 div.hoot-login-container p+h2 {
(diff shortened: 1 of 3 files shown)


brianhatchl:
Is there some way to test this fix?

I did notice that if I logged out on branch `1242`, the app remembered my username when showing the OAuth popup.  This makes it impossible to change users.  I did try deleting the `user` key from local storage, but no joy.

sumarlidason:
let me work this to generate scenarios --

sumarlidason:
These changes improve the browser's reaction to failure to authenticate in the final step of the workflow: If the user comes back to us with an invalid token or a token that has expired. 

* ability for the child window to pass back an error to the parent window
`window.oAuthDone = function(user_object) {`
to 
`window.oAuthDone = function(e, user_object) {`
* better error handling in response to `/hoot-services/auth/oauth1/verify` failures

We can recreate this issue by 
* changing OAuth1Resource.java:73 from 'verify' to 'verif'
* redeploy
* logout
* attempt to login

On failure, with current logic, the user is left with this result:
screenshot from 2018-10-23 12-59-17

this touches the the following code,
screenshot from 2018-10-23 13-11-29

Issue #1248, opened 2018-10-30
update user URIs


sumarlidason:
* replace with user-aware `getCurrentUser`

sumarlidason:
https://github.com/ngageoint/hootenanny/issues/2590

PR #1251, opened 2018-11-01
update user URIs

sumarlidason:
https://github.com/ngageoint/hootenanny-ui/issues/1248
https://github.com/ngageoint/hootenanny/issues/2590


sumarlidason:
application critical URIs modified, see https://github.com/ngageoint/hootenanny/pull/2724

PR #1268, opened 2018-11-20
Hoot2x oauth integration

putipong:
#### Express
- Proxying is now used to handle all outgoing network requests and will translate `localhost` to the remote server's address
- Middleware is used in Karma tests to spoof a logged-in user by adding a `SESSION` cookie to all request headers
- Dev server port has changed from `9000` to `8080`. This is to match the schema which hoot-services follows

#### Codesplit
_Production_ Webpack bundles are now chunked into smaller, more concise files. Each page has its own primary bundle, as well as `vendor` bundles that contain all 3rd party modules (lodash, d3, etc.) used by the primary bundle. Some vendor files contain shared modules between pages, in which case it will be delimited with `~`. ex:
- `vendors~iD~login.min.js` contains common modules shared between iD and login
- `vendors~iD.min.js` contains modules used only in iD

#### Navbar Dropdown
The "About" icon on the right side of the navbar is now a dropdown list that can be toggled. It shows the current logged in user and a few other clickable options. Once logged in, this is where users can view info about the app or log out of their session. Logging out will automatically redirect back to the login page.

## Installation
Run `npm i` after pulling down project to make sure you have the latest versions of modules

#### Development
- `npm start`
- go to `http://localhost:8080`

#### Production
Again, proxying all requests from localhost to hoot-services so you'll need to start an express server.
- `npm run production`
- `cd hoot-server`
- `npm i` (if needed)
- `npm start`
- go to `http://localhost:8080`

#### Test
- `npm run test:hoot`

After Webpack finishes bundling, Chrome should automatically open and begin executing the tests. Code coverage summary will output to CLI when tests are complete and full details can be viewed by going to `/coverage/index.html`.


dist/index.html
```diff
@@ -39,39 +39,45 @@
...
+                                'oauth_consumer_key': 'zwQZFivccHkLs3a8Rq5CoS412fE5aPCXDw9DZj7R',
+                                'oauth_secret': 'aMnOOCwExO2XYtRVWJ1bI9QOdqh1cay2UgpbhA6p'
+                            }
+                            ])
+                        );
+                });
+            }
+        } )
+        .catch( () => { // user not logged in - redirect to login page
+            window.location.replace( '/login.html' );
```
brianhatchl: Is this going to work when the UI is not at the root, but under `/hootenanny-id` when deployed with tomcat?

modules/Hoot/config/apiConfig.js
```diff
@@ -5,13 +5,13 @@
...
-  mapnikServerPort: "8000",
-  mergeServerPort: "8096",
-  queryInterval: 2000
+    host: 'http://localhost',
+    port: '8080',
+    path: 'hoot-services',
+    translationServerPort: '8094',
+    mapnikServerPort: '8000',
+    mergeServerPort: '8096',
+    queryInterval: 2000
```
brianhatchl: Will this need updating per deployed environment?

modules/Hoot/hoot.js
```diff
@@ -26,7 +30,7 @@ class Hoot {
         this.events       = new EventManager();
 
         this.config = {
-            urlroot: 'http://54.165.118.78:8080/hoot-services/osm',
+            urlroot: 'http://localhost:8080/hoot-services/osm',
```
brianhatchl: Will this need updating per deployed environment?

modules/Hoot/managers/api.js
```diff
@@ -446,7 +487,12 @@ export default class API {
         };
 
         return this.request( params )
-            .then( resp => resp.data );
+            .then( async resp => {
+                await setTimeout( () => {
+                    console.log( 'blahhh' );
+                }, 30000 );
```
brianhatchl: Remove this?

modules/Hoot/ui/navbar.js
```diff
@@ -44,49 +46,83 @@ export default class Navbar {
...
+
+            let dropdownContent = rightContainer
+                .append( 'ul' )
+                .classed( 'dropdown-content fill-white', true );
+
+            dropdownContent
+                .append( 'li' )
+                .classed( 'dropdown-item pad2x strong', true )
+                .append( 'span' )
+                .text( 'Logged in as ' + user.display_name );
```
brianhatchl: I personally like the display name shown in the nav bar like Hoot v1.  Not sure if that opinion is shared by anyone else.

modules/renderer/background.js
```diff
@@ -175,9 +175,9 @@ export function rendererBackground(context) {
             delete q.offset;
         }
 
-        if (!window.mocha) {
-            window.location.replace('#' + utilQsString(q, true));
-        }
+        // if (!window.mocha) {
+        //     window.location.replace('#' + utilQsString(q, true));
+        // }
```
brianhatchl: I think we can re-enable this after splitting login.html from index.html.

dist/login.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset='utf-8'>
+    <title>Hootenanny</title>
+    <base href="/">
```
brianhatchl: I think this tag is breaking deployment under tomcat `/hootenanny-id2x`, because we can't assume all resources are at the root.  I'm doing a test deploy to see what else arises.

modules/Hoot/login.js
```diff
@@ -0,0 +1,282 @@
...
+
+        row = table
+            .append( 'div' )
+            .classed( 'hoot-login-row', true );
+
+        row
+            .append( 'div' )
+            .classed( 'hoot-login-cell', true )
+            .append( 'img' )
+            .attr( 'src', '/img/login-popup-firefox-1.png' );
```
brianhatchl: This path needs to be made relative `./img`

modules/Hoot/ui/login.js
```diff
@@ -0,0 +1,226 @@
...
+
+        row = table
+            .append( 'div' )
+            .classed( 'hoot-login-row', true );
+
+        row
+            .append( 'div' )
+            .classed( 'hoot-login-cell', true )
+            .append( 'img' )
+            .attr( 'src', '/img/login-popup-firefox-1.png' );
```
brianhatchl: This path needs to be made relative `./img`

dist/login.html
```diff
@@ -0,0 +1,36 @@
...
+    <link rel="icon" href="favicon.ico"/>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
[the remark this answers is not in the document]
putipong: yes that can be done

dist/index.html
```diff
@@ -19,8 +18,9 @@
     <div id="id-sink"></div>
 </div>
 <!--[if !IE]>-->
-<script src='./hoot.min.js'></script>
 <script src='./iD.min.js'></script>
+<script src='./vendors~iD.min.js'></script>
+<script src='./vendors~iD~login.min.js'></script>
```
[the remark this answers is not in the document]
putipong: correct – they are only needed in `production`. `develop` will run fine without them, though will give a `404` if they don't already exist. i'll look into a way to inject the script tag when building for prod

dist/index.html
putipong: good catch. i'll update it so that only the last part of the path changes

modules/Hoot/config/apiConfig.js
putipong: AFAIK, this should be consistent with the ports in production

modules/Hoot/hoot.js
putipong: this should be fine in all environments, including testing

modules/Hoot/ui/navbar.js
putipong: subject to change. it wouldn't take long to move back out to the navbar, but i think it took up a lot of unnecessary space. maybe it can display in the navbar briefly when the user logs in or opens the page, and then fades out?

modules/Hoot/config/apiConfig.js
brianhatchl: Could this use `window.location.host` and `window.location.port` instead?

putipong: Unless you're talking about the UI, that wouldn't work because the translation, mapnik, and merge servers run on different port from the UI.

putipong: Also, the api config is bundled alongside the rest of the code so we'd have to update ports during the bundling process. This can be done with node environment variables

brianhatchl: Just for the protocol, hostname, port to hit the UI.

dist/index.html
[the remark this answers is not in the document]
brianhatchl: ok, benign 404s are fine in development, but caught my eye.

modules/Hoot/ui/navbar.js
brianhatchl: I'm thinking of testing with users.  I now have two users and have a habit of glancing up at who I'm logged in as to make sure I'm seeing the correct folders/datasets.

modules/Hoot/hoot.js
brianhatchl: I don't see how this will work in production, when Hoot is not running or proxied locally.

putipong:
Fixed your change requests @brianhatchl
- `/dist/img/` has been deleted and added it to `.gitignore` so images aren't duplicated in the repo
- Page redirection now uses relative path for cross-env compatibility
- Now using `window.location` in `apiConfig.js` for cross-env compatibility

brianhatchl:
:shipit:

PR #1280, opened 2018-12-04
Translation assistant fix

putipong:
- Added borders to TA mapping preset rows for more clarity between rows
- Fixed bug that caused mapping UI to render multiple times when re-uploading a translation

#### Testing instructions
With translation assistant panel open, upload a translation file (only .zip works at the moment). After upload is complete and the mapping UI has rendered, re-upload the same file. As a result, the mapping UI will be destroyed and re-rendered. _A duplicate should not be appended to the container_.


brianhatchl:
A couple more things:
* Can the cursor override all elements in the div?  I see it when the cursor is not over buttons, but would better match hoot1 if the wait cursor was shown no matter what was hovered in the pane.
* Getting this error when switching data layer in the drop down:
```
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?
    at HTMLFormElement.remove (http://localhost:8080/iD.min.js:103382:22)
    at Selection.each (http://localhost:8080/iD.min.js:102763:37)
    at Selection.remove (http://localhost:8080/iD.min.js:103386:15)
    at TagMapForm.changeLayer (http://localhost:8080/iD.min.js:8320:30)
    at HTMLInputElement.<anonymous> (http://localhost:8080/iD.min.js:8361:26)
    at HTMLInputElement.<anonymous> (http://localhost:8080/iD.min.js:103199:16)
    at HTMLInputElement.<anonymous> (http://localhost:8080/iD.min.js:72359:14)
    at Selection.each (http://localhost:8080/iD.min.js:102763:37)
    at utilTriggerEvent (http://localhost:8080/iD.min.js:72356:12)
    at HTMLAnchorElement.accept (http://localhost:8080/iD.min.js:25664:83)
```

putipong:
@brianhatchl fixed mentioned bugs

brianhatchl:
Those two issues are fixed, but in validating them I found one more issue.  The old TA would allow the user to press <kbd>Enter</kbd> and the typed text from the dropdown would be chosen.

In this version, <kbd>Enter</kbd> caused a new Search Tag box to appear.

putipong:
@brianhatchl I made changes that should resolve the bug pointed out

brianhatchl:
@putipong I think we want Enter to call this 
https://github.com/ngageoint/hootenanny-ui/blob/7b8f618194671374a8975bae716833f9c7404bf0/modules/Hoot/ui/managePanel/transAssist/tagMapWidget.js#L129-L133
rather than fire change.

putipong:
@brianhatchl Sorry, I misunderstood which dropdown you were talking about. I thought you meant the one for layer selection

putipong:
updated to allow scrolling, accepting, and canceling on the tag list

brianhatchl:
I do really like the ability to down arrow and scroll through the suggested values.  But we also need the ability for a user to type their own value (that's not in the suggested list) and use that.
In Hoot 1 that was triggered by typing and then pressing Enter, to accept the text as-is in the box.
Does that make sense?  You can try it out in Hoot 1 to see what I mean.

putipong:
@brianhatchl Okay I can change it to allow users to type a custom tag value

brianhatchl:
LGTM :owl:

PR #1290, opened 2018-12-11
1264

jackgrossman18:
Added requested test for conflate button interaction


putipong (changes requested):
You got the idea of creating mock data needed for tests down, but you're also repeating tests that have already been implemented in other suites. It's best to avoid this because it makes it more difficult to track down what went wrong if one or the other fails and it causes tests to run longer than needed.

Instead, try manually uploading data by directly invoking the API with params in a `before()` function. Doing this will ensure that you have all the test files you need _before_ the actual tests begin.

PR #1298, opened 2019-01-09
Add "Attribute" conflation type and adv option

brianhatchl:
for "Remove unmatchedRemove Unmatched Secondary Features"


brianhatchl:
This is not quite ready yet.  The individual adv option works, but the Attribute conflation type doesn't seem to get it's adv option overrides applied.

Issue #1301, opened 2019-01-10
Remove Average Conflation method from User Interface


davidkulbeth:
Decision made after discussion with our VGI teams in SFD and SFG.

brianhatchl:
@jackgrossman18 I made this change for Hoot1, but you can make it for Hoot2x.

PR #1303, opened 2019-01-10
remove Average conflation type

brianhatchl:
and also generate report artifacts

refs #1301


brianhatchl:
This PR will rely on https://github.com/ngageoint/hootenanny/pull/2847 to run tests because a cuke test for average conflation needs removing (actually modifying to make an attribute conflation test).

Issue #1308, opened 2019-01-11
Attribute conflation: default attribute reference layer to secondary


brianhatchl:
The intent of Attribute conflation is to override attributes of a reference layer (having better geometry) with tags of the secondary layer (having more descriptive attributes).

The option of `Attribute reference layer` is one of the conflation controls, but should be set to Secondary by default for Attribute conflation.

brianhatchl:
@jackgrossman18 I'll make this change for Hoot1 and you can do Hoot2x.

brianhatchl:
@jackgrossman18 make your branch name `1308_2x` plz

PR #1310, opened 2019-01-11
Default attribute reference to secondary layer for Attribute conflation

brianhatchl:
refs #1308


brianhatchl:
Choosing `Attribute` conflation type should change the Attribute Reference Layer to the name of the secondary layer.  Changing it to any other conflation type to set it to the reference layer name.

image

PR #1313, opened 2019-01-15
add new conflation opts that don't override defaults

brianhatchl:


brianhatchl:
Took a stab at getting the new (not overridden) conflation opts to add to the `ADV_OPTIONS` job param.

PR #1327, opened 2019-02-01
try/catch import buildInfo.json

putipong:


brianhatchl:
I will see if buildInfo gets copied by the services makefile before the ui2x build runs

brianhatchl:
So my centos vm was having issues just getting all the node module dependencies installed, so I didn't get a good test of the above.

@putipong is there a way to add a `git describe --tags` step to the webpack build that could generate the build info file for the UI repo?

brianhatchl:
@sputnam I can't build hoot-2x on the centos 7 vagrant vm because it doesn't have node 10.
I'm assuming the jenkins vm has this available?

brianhatchl:
>way to add a git describe --tags step to the webpack build that could generate the build info file for the UI repo

Added an approach for this here https://github.com/ngageoint/hootenanny-ui/pull/1329/commits/64f73de91d0adc1408a5dadea408195df9849ce5

This does result in a different version for the UI, but it is actually accurate.

PR #1337, opened 2019-02-12
show error message when trying to delete other user's resource #1336

maxgrossman:
closes #1336

### DIFF
--- a/modules/Hoot/managers/api.js
+++ b/modules/Hoot/managers/api.js
@@ -178,7 +178,7 @@ export default class API {
      */
     getJobStatus( id ) {
         const params = {
-            path: `/job/status/${ id }`,
+            path: `/job/status/${ id }?includeCommandDetail=true`,
             method: 'GET'
         };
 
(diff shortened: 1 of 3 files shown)


modules/Hoot/ui/managePanel/datasets.js
```diff
@@ -194,10 +194,20 @@ export default class Datasets extends Tab {
...
+                        });
                 } else {
                     return Hoot.api.deleteFolder( data.id )
-                        .then( () => Hoot.folders.removeFolder( data.id ) );
+                        .then( () => Hoot.folders.removeFolder( data.id ) )
+                        .catch( ( err ) => {
+                            err.message = err.data;
+                            delete err.data;
+                            Hoot.message.alert( err );
+                        });
```
brianhatchl: Does this effectively remove the `More details` link from success alerts?

modules/Hoot/managers/api.js
```diff
@@ -564,7 +564,9 @@ export default class API {
             } )
             .catch( err => {
                 return Promise.reject( {
-                    data: err.data,
+                    data: {
+                        details: err.data.commandDetail[0].stderr
```
brianhatchl: Should we do the right thing and do a `join(';')` or something instead of `[0]`?

PR #1348, opened 2019-02-14
restore buildInfo.json to fix rpm builds

brianhatchl:
will have to be updated manually with 'npm run buildinfo'


jbronn:
thanks @brianhatchl!

PR #1352, opened 2019-02-18
adding coords back to footer in ui

jackgrossman18:
@brianhatchl 

adding coords back to ui within the footer

ref #1318


modules/ui/init.js
```diff
@@ -249,6 +249,11 @@ export function uiInit(context) {
                 .call(uiAccount(context));
         }
 
+        aboutList
+            .append('li')
+            .attr('class', 'info-block')
+            .call(uiCoordinates(context));
+
```
brianhatchl: Looking at the bottom bar in the UI, I think the coordinate display would be more at home next to the scale bar on the left.  It also has the behavior of changing units when clicked.  I do like this better than the right-click menu, but please add a tool tip like in Hoot1, eg. "Click to change coordinate format".  Also I think MGRS should be one of the supported formats.

modules/ui/coordinates.js
```diff
@@ -143,8 +148,17 @@ export function uiCoordinates(context) {
         return zone.toString() + sn + ' ' + Xutm.toString() + 'm E ' + Yutm.toString() + 'm N';
     }
 
+    function invaildCoordinates(coords) {
```
brianhatchl: typo in "invalid"

jackgrossman18:
@brianhatchl moved the coords to the left of the footer, added a popup window like hoot 1, and included MGRS as an option

brianhatchl:
Did some more clean up as seen in commit messages

PR #1355, opened 2019-02-22
Install hoot from RPM

sputnam:
Instead of compiling hootenanny to test the UI, we could install the latest RPM in the develop repository.  Once installed run UI tests against hoot install.  In theory, this should save time and cut down on unnecessary compiling.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,8 +40,10 @@ pipeline {
                 expression { return params.UI }
             }
             steps {
-                sh "vagrant ssh ${params.Box} -c 'cd hoot; source ./SetupEnv.sh; aclocal && autoconf && autoheader && automake --add-missing --copy && ./configure --quiet --with-uitests'"
-                sh "vagrant ssh ${params.Box} -c 'cd hoot; source ./SetupEnv.sh; make ui2x-build; time -p make -s ui2x-test'"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -27,15 +27,23 @@ pipeline {
         stage("Vagrant Up") {
             steps {
                 // TODO: Vagrant up --noprovision, install hoot from daily develop RPMs
-                sh "vagrant up ${params.Box} --provider aws"
+                sh "vagrant up ${params.Box} --provider aws --provision-with hoot"
```
brianhatchl: Would we even need to provision this box at all?  The RPM should install all dependencies on a bare centos 7 box.

sputnam: @brianhatchl The UI tests require at least chrome and that doesn't come from the hoot RPM so there is some level of provisioning still required even after the hoot RPM install (I believe).  I'll try a few permutations locally and see if I can come up with the minimum provisioning required.

brianhatchl: Ok, true.  My concern is that the `hoot` provision does quite a lot, and is probably duplicative of what the rpm installs (pg, tomcat, etc).
Maybe we need a hoot-ui-test rpm package?

Issue #1328, opened 2019-02-01
Unable to download translations, js error


brianhatchl:
```
Uncaught (in promise) Error: TypeError: Object(...) is not a function
    at Translations.exportTranslation (translations.js:193)
```

PR #1361, opened 2019-03-04
translations now download

jackgrossman18:
ref #1328


jackgrossman18:
@brianhatchl selected translations from the translations modal now download files properly

PR #1380, opened 2019-03-08
New lint rules for no trailing spaces and no tabs

brianhatchl:


brianhatchl:
:bulb: Recommend configuring your editor to remove trailing spaces and convert tabs to spaces on save.

Issue #1372, opened 2019-03-07
Saving New Translation, from Translation Assistant 


kweint:
After completing translation assistant, the save button returns the following window:

![image](https://user-images.githubusercontent.com/18521019/53994335-120a9680-4100-11e9-9471-69c8add88cea.png)

Most notably, the translation js is not present.

jackgrossman18:
@kweint  did you experience this after you added this new translation?

If not, it looks like the reason isn't saving is because no translation text was put into the translation box

brianhatchl:
So the translation assistant should auto-populate the text area with the generated translation file.

Check Hoot1 for it's behavior.

PR #1387, opened 2019-03-18
translation now previewed within new translation pop-up

jackgrossman18:
ref #1372


jackgrossman18:
@brianhatchl when uploading a new translation, the subsequent popup shows the translation.js file appropriately within the dialog

PR #1389, opened 2019-03-21
1385 2x. Clip will pass in folder id

bobi7:
this will allow the clip endpoint to be able to move the clipped item to the specified directory


modules/Hoot/tools/clipDataset.js
```diff
@@ -208,16 +210,23 @@ export default class ClipDataset {
             let row         = d3.select( `#row-${ mapId }` ),
                 datasetName = row.select( '.datasetName' ),
                 outputName  = row.select( '.outputName' ),
-                pathName    = row.select( '.outputPath' );
+                pathName    = row.select( '.outputPath' ).property( 'value' ),
```
brianhatchl: I feel like the folders when chosen from the UI should set their id, rather than have to lookup by name.

modules/Hoot/managers/api.js
```diff
@@ -936,7 +936,27 @@ export default class API {
...
+                    type: resp.type
+                };
+            } )
+            .catch( err => {
+                let message, status, type;
+
+                status = err.status;
+
+                if ( status >= 500 ) {
+                    message = 'Error during Clip! Please try again later.';
```
brianhatchl: we should use the technique from the jobs panel branch (1331) to get the `stderr` value here

PR #1393, opened 2019-03-28
export and job percent bug fix

brianhatchl:


modules/Hoot/ui/modals/exportData.js
```diff
@@ -39,7 +39,7 @@ export default class ExportData {
         this.appendToFgdbCheckbox = this.container.select( '#exportAppendFgdb' );
         this.dataExportNameTextInput = this.container.select( '#dataExportNameTextInput' );
         this.submitButton = this.container.select( '#exportDatasetBtn' );
-        this.submitButton.attr('disabled', false);
+        this.submitButton.attr('disabled', null);
```
brianhatchl: This clears the attribute, instead of changing it's value.  Makes export data dialog button enabled with defaults.

test/index.html
```diff
@@ -136,7 +136,7 @@
 
     <script src='spec/ui/fields/access.js'></script>
     <script src='spec/ui/fields/localized.js'></script>
-    <script src='spec/ui/fields/wikipedia.js'></script>
+    <!--script src='spec/ui/fields/wikipedia.js'></script-->
```
brianhatchl: This iD test was constantly failing.  Think it relies on remote service so disabling.

modules/Hoot/ui/modals/exportData.js
```diff
@@ -14,8 +14,8 @@ export default class ExportData {
     constructor( translations, d, type ) {
         const isDatasets = type === 'Datasets';
         this.translations = translations;
-        this.input = isDatasets ? d.map(n => n.data.name).join(',') : d.data.name;
-        this.id = isDatasets ? d.map(n => n.data.id).join(',') : d.data.id;
```
brianhatchl: `data` is not populated for multiple selected datasets

PR #1400, opened 2019-04-05
replace uglify with terser for production build

brianhatchl:
UglifyJs started returning this error during build on centos vm
```
ERROR in vendors~iD.min.js from UglifyJs
Unexpected token: keyword «const» [vendors~iD.min.js:4193,2
```


brianhatchl:
Adding the `package-lock.json` file back to git (the version generated on the CentOS 7 vagrant vm) seems to have fixed the hoot2x UglifyJs build error.

PR #1401, opened 2019-04-08
restore tracking of package-log.json

brianhatchl:
the centos 7 file generated differed from
the ubuntu 16 file generated when using
node 8

the differences could account for the failing hoot2x builds?


brianhatchl:
The file committed is the `package-lock.json` file for the centos 7 vagrant vm.

PR #1402, opened 2019-04-09
grail ui tools added

bobi7:


brianhatchl:
Need to update test expectation for `hoot-ui/test/hoot/spec/toolbar/tools.js`

brianhatchl:
@bobi7 can you post a screenshot of the new changeset summary table?

bobi7:
Screen Shot 2019-05-31 at 11 09 41 AM

brianhatchl:
Ah, do we need to show another table or update the numbers if Apply Tag Differential is checked?

Also, that check does not appear if the changeset is the the result of a normal conflation right?

bobi7:
correct normal conflation in UI doesn't make a tags osc file so no check. Derive differential changeset has to be run for that to occur

PR #1403, opened 2019-04-10
Conflate: inherit output folder from inputs, add cancel job button, improve error msgs

brianhatchl:
Also sends folder id so output can be moved without subsequent http request.


maxgrossman:
ref #1396

maxgrossman:
ref #1399

Issue #1377, opened 2019-03-08
UI allows a user to upload a map dataset with an existing name and overwrites it


brianhatchl:
Hoot1 UI checks for duplicate map names and disallows this.  In theory duplicate names should be allowed once all our map references use map id and not map name.  But overwriting an existing map seems wrong.

PR #1413, opened 2019-04-18
added check for layer name, if it exists, append a number to end of name

jackgrossman18:
ref #1377

### DIFF
--- a/modules/Hoot/ui/modals/importDataset.js
+++ b/modules/Hoot/ui/modals/importDataset.js
@@ -346,7 +346,7 @@ export default class ImportDataset {
             NONE_TRANSLATION: translation.NONE === 'true',
             TRANSLATION: translationName,
             INPUT_TYPE: importType.value,
-            INPUT_NAME: this.getLayerManager( layerName ),
+            INPUT_NAME: Hoot.layers.checkLayerName( layerName ),
             formData: this.getFormData( this.fileIngest.node().files )
         };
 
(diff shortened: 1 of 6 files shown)


modules/Hoot/ui/modals/importDataset.js
```diff
@@ -363,6 +363,16 @@ export default class ImportDataset {
                 Hoot.events.emit( 'modal-closed' );
             } );
     }
+    checkLayerName(layerName) {
```
brianhatchl: I would move this code to layerManager.js.

I think we also want to use it when doing any new dataset output (conflation & clipping as well as import).
Maybe we could replace the UUID suffix that doesn't give a good sense of which output is most recent.

I think clipping output names just add a number to the end, so maybe we can consolidate the name conflict resolution behaviors so they are all in common (`OutputName_uuid` vs `OutputName_1` vs `OutputName1` could all become `OutputName (1)`)

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -363,6 +363,16 @@ export default class ImportDataset {
...
+        var matches = [];
+        for (let i = 0; i < Hoot.layers.allLayers.length; i++) {
+            var checkedLayer = Hoot.layers.allLayers[i].name;
+            if (checkedLayer === layerName || new RegExp(layerName + '_\\d+').test(checkedLayer) ) {
```
brianhatchl: The `_\\d+` regex here also matches with the generated UUID suffixes used by conflation.  Could we make it `OutputName (2)` like browser downloads do when a filename already exists?

jackgrossman18:
@brianhatchl all requested changes have been made

Issue #1408, opened 2019-04-12
add visibility toggle for the datasets


bobi7:
same functionality as in hoot 1 ui

PR #1414, opened 2019-04-19
1408. added dataset toggle + button to switch which layer is on top

bobi7:
refs #1408

### DIFF
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -348,11 +348,10 @@ export default class Layers {
               id = layer.id;
 
         if (isVisible) {
-            this.hoot.context.connection().addTile( id );
+            d3.selectAll(`.tag-hoot-${id}`).attr('display','');
         } else {
-            this.hoot.context.connection().removeTile( id );
+            d3.selectAll(`.tag-hoot-${id}`).attr('display','none');
         }
-        this.hoot.context.flush();
     }
 
     decodeHootStatus( status ) {
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -43,6 +43,7 @@ export default class Layers {
         this.mergedLayer        = null;
         this.mergedConflicts    = null;
         this.palette            = colorPalette;
+        this.topLayer           = null;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


modules/ui/background.js
```diff
@@ -283,6 +288,77 @@ export function uiBackground(context) {
...
+            .merge(updateList);
+
+        let updateButton = updateList.selectAll('.lyrcntrlbtn')
+            .data(d => [d]);
+
+        updateButton.exit().remove();
+
+        updateButton = updateButton.enter()
+            .append('button')
+            .call(tooltip().title('Click to have layer be on top').placement('left'))
```
maxgrossman: sick! Though I think we don't want to have this for the top most layer. I think we can use the `.each()` here method which gives us the index so we can ignore the 0th....

I think it's something like

```javascript
.each(function(d, i) {
     if (i===0) return
     d3.select(this).call(/*the tooltip stuff you have*/)
}

File: modules/Hoot/managers/layerManager.js [hunk 2]
```diff
@@ -342,6 +343,18 @@ export default class Layers {
         this.hoot.context.flush();
     }
 
+    toggleLayerVisibility( layer ) {
+        const isVisible = layer.visible = !layer.visible,
+              id = layer.id;
+
+        if (isVisible) {
+            this.hoot.context.connection().addTile( id );
```
maxgrossman: I noticed that these actions cause a sort of flicker to occur when I toggle on and off. 
I think we can avoid this be just switching layer opacity between 0 and 1 for the layers' svg path elements...
Looking at the html for one of the paths...

```html
<path class="way line stroke w4_273 tag-highway tag-highway-road tag-hoot-273" d="M917.0692347958684,89.5553780272603L1010.1036382615566,90.6621542610228"></path>
```

I think that we can do a selectAll in the `.data-layer .osm` g element, then select all those with the `tag-hoot-273` classes in this example, then change opacity

modules/Hoot/managers/layerManager.js
```diff
@@ -377,6 +390,15 @@ export default class Layers {
         sheets.insertRule( 'g.point.tag-hoot-' + mapId + ' .stroke { fill:' + color + '}', sheets.cssRules.length - 1 );
     }
 
+    setTopLayer( mapId ) {
+        this.topLayer = mapId;
+        this.hoot.context.flush();
```
maxgrossman: is this what you call to get the additions you put in the lines.js file to occur?

bobi7: correct. ive seen this flush get called and it always seems like a 'ok initialize the ui with this newly edited/added info'

modules/Hoot/managers/layerManager.js
bobi7: yea i tried this method as well but noticed the click handlers remained on the map. Will look into way to remove those because this approach is more efficient and none of the dumb flicker

modules/services/osm.js
```diff
@@ -1241,6 +1241,11 @@ export default {
         dispatch.call( 'loaded' );
     },
 
+    addTile( id ) {
```
maxgrossman: does this get used?

bobi7: good call. removed

PR #1417, opened 2019-04-19
GeoJSON now an upload option w/in Import Muti Dataset modal

jackgrossman18:
ref #1416


jackgrossman18:
@brianhatchl removed code the crept over from separate PR

Issue #1420, opened 2019-04-23
loadTiles making erroneous request at zoom level 1


brianhatchl:
Some shenanigans going on here:
https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/renderer/map.js#L618-L643

PR #1423, opened 2019-04-23
fix tile id to use both id and map id

brianhatchl:
refs #1420

also reverts some map code to be more like upstream iD

### DIFF
--- a/modules/services/osm.js
+++ b/modules/services/osm.js
@@ -987,8 +987,9 @@ export default {
                 .zoomExtent([_tileZoom, _tileZoom])
                 .getTiles(projection)
                 .map(function(tile) {
-                    tile.mapId = tile.id = layer.id;
+                    tile.mapId = layer.id;
                     tile.layerName = layer.name;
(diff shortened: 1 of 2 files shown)


modules/services/osm.js
```diff
@@ -987,8 +987,9 @@ export default {
                 .zoomExtent([_tileZoom, _tileZoom])
                 .getTiles(projection)
                 .map(function(tile) {
-                    tile.mapId = tile.id = layer.id;
+                    tile.mapId = layer.id;
                     tile.layerName = layer.name;
+                    tile.id = tile.id + '_' + tile.mapId;
```
brianhatchl: This was the crux of the bug.  The tile id is used as the key in the tile cache.  The old code just used the map id which obviously will be one entry per dataset layer.  We need the key to be a combo of the tile id and the map id.

A lot of the other code I removed was a workaround for this bug.

brianhatchl: And reverted to restore upstream iD lines of code

Issue #1425, opened 2019-04-25
adding new features when 2 datasets are loaded doesnt work


bobi7:
add 2 datasets
add a new line connected to one of the datasets
save and you'll see "Changeset already inflight"

PR #1427, opened 2019-05-08
Resolving Changeset already inflight error when editing with 2 layers w/in map

jackgrossman18:
ref #1425


modules/Hoot/managers/api.js
```diff
@@ -327,6 +327,32 @@ export default class API {
             } );
     }
 
+
+    /**
+     * Get all id's necessary to create new map features
+     *
+     * @param {Promise|array} mapId + changesetId, nodeId, wayId, relationId
```
maxgrossman: Just a reminder, let's update this comment!

modules/Hoot/managers/api.js
```diff
@@ -327,6 +327,32 @@ export default class API {
...
+    getAllIds( mapId ) {
+        const params = {
+            path: `/osm/api/0.6/map/${ mapId }/startingIndex`,
+            method: 'GET'
+        };
+
+        return this.request( params )
+            .then( resp =>  {
+
+                let activeIds = {};
```
maxgrossman: I think just return resp.data!

modules/Hoot/ui/sidebar/sidebar.js
```diff
@@ -15,6 +15,7 @@ import {
     utilQsString,
     utilStringQs
 } from '../../../util';
+import { diff3MergeIndices } from 'node-diff3';
```
maxgrossman: think we can remove this import!

modules/behavior/draw_way.js
```diff
@@ -73,6 +74,8 @@ export function behaviorDrawWay(context, wayId, index, mode, startGraph) {
     function move(datum) {
         context.surface().classed('nope-disabled', d3_event.altKey);
 
+    // add one more boolean at end, is activeLayer===mapId
+        var _activeLayer = _find( Hoot.layers.loadedLayers, function(a, b) { return a.activeLayer; });
```
maxgrossman: I think we need to make use of this here, making sure to only apply a target loc if active ids match.

(here in iD)[https://github.com/openstreetmap/iD/blob/master/modules/behavior/draw_way.js#L85]  a question is asked (is this target loc one that allows vertex).

Our question needs to be `does the datum's mapId equal the _activeLayer mapId`

modules/osm/entity.js
```diff
@@ -3,40 +3,51 @@ import _keys from 'lodash-es/keys';
...
 
     // Create the appropriate subtype.
     if (attrs && attrs.type) {
-        return osmEntity[attrs.type].apply(this, arguments);
-    } else if (attrs && attrs.id) {
-        return osmEntity[osmEntity.id.type(attrs.id)].apply(this, arguments);
-    }
 
+        var _activeLayer = _find(Hoot.layers.loadedLayers, function(a, b) { return a.activeLayer; });
+        if (_activeLayer && _activeLayer.id) {
```
maxgrossman: totally just a stylistic suggestion, so no need to change, but I almost wonder if we should make a new func for grabbing the ids from the active layer. something like `osmEntity.id.hoot` that looks a little like this...

```javascript
function(type, activeLayer) {
   return osmEntity.id.fromOsm(type, activeLayer.activeIds[type]--) + '_' + activeLayer.id;
}
```

Then here in this if block you just say "do I have an activeLayer'?
Ok I do, use `id.hoot` function
If I don't use the out of the box id way of doing things.

just a thought!

jackgrossman18:
@maxgrossman please take a look at the most recent changes

modules/Hoot/ui/sidebar/sidebar.js
```diff
@@ -168,6 +169,116 @@ export default class Sidebar {
...
+        let loadedLayers    = Object.values(Hoot.layers.loadedLayers);
+        let selectReference = d3.selectAll('#reference');
+        let selectSecondary = d3.selectAll('#secondary');
+        if (loadedLayers.length === 2) {
+            let referenceActive = _find(loadedLayers, function(a, b) { return a.refType === 'primary'; });
+            let secondaryActive = _find(loadedLayers, function(a, b) { return a.refType === 'secondary'; });
+            let changeActive    = new LayerAdd();
+
+
+            if (d3.select('#reference button.select-active-layer').empty()) {
```
maxgrossman: wonder if the on click callback can be generalized in a single function? seems like when we click one fo the active layer buttons, we make the one we click's classes all set to active, and the other one's set to false. so a function could have 2 params, one we want to set to active, one we want to disable and just use those accordingly.

so naively, something like

```javascript
function update(active, inactive) {
   //...update active button class
   //...update inactive button
}
```

modules/behavior/draw_way.js
```diff
@@ -279,6 +287,8 @@ export function behaviorDrawWay(context, wayId, index, mode, startGraph) {
 
     // Connect the way to an existing node and continue drawing.
     drawWay.addNode = function(node, d) {
+        console.log(node);
```
maxgrossman: let's remove the stray console logs!

modules/behavior/draw_way.js
```diff
@@ -73,7 +75,9 @@ export function behaviorDrawWay(context, wayId, index, mode, startGraph) {
     function move(datum) {
         context.surface().classed('nope-disabled', d3_event.altKey);
 
+        var _activeLayer = _find( Hoot.layers.loadedLayers, function(a,b) { return a.activeLayer; });
         var targetLoc = datum && datum.properties && datum.properties.entity && datum.properties.entity.loc;
```
maxgrossman: I think we need to implement the commented out piece here, using the same split stuff we do in the draw.js file.

modules/behavior/hover.js
```diff
@@ -140,6 +140,7 @@ export function behaviorHover(context) {
                     return;
                 }
 
+				// again, boolean, is entity.mapId === activeLayerId...
                 var suppressed = _altDisables && d3_event && d3_event.altKey;
```
maxgrossman: same is activeLayer check here!

modules/osm/entity.js
```diff
@@ -3,40 +3,45 @@ import _keys from 'lodash-es/keys';
...
     return (new osmEntity()).initialize(arguments);
 }
 
-
 osmEntity.id = function(type) {
-    return osmEntity.id.fromOSM(type, osmEntity.id.next[type]--);
+    var _activeLayer = _find(Hoot.layers.loadedLayers, function(a, b) { return a.activeLayer; });
+    let allIds = {};
+    allIds = _activeLayer.activeIds;
+    var id = osmEntity.id.fromOSM(type, allIds[type]--);
```
maxgrossman: I think you can call the `_activeLayer.activeIds` directly inside the fromOsm function `_activeLayer.activeIds[type]--`

modules/services/osm.js
```diff
@@ -41,6 +42,7 @@ import {
 } from '../util';
 
 import { baseUrl as hootBaseUrl } from '../Hoot/config/apiConfig';
+import { all } from 'q';
```
maxgrossman: import to remove I do believe!

modules/services/osm.js
```diff
@@ -8,6 +8,7 @@ import _flatten from 'lodash-es/flatten';
 import _groupBy from 'lodash-es/groupBy';
 import _isEmpty from 'lodash-es/isEmpty';
 import _map from 'lodash-es/map';
+import _matches from 'lodash-es/matches';
```
maxgrossman: import to remove!

jackgrossman18:
@maxgrossman please take a look at the new changes

maxgrossman:
The changeset upload logic is :man_cook: :kiss: - made changes to both the secondary and reference datasets, then checked their respective changesets/nodes tables. they had the right changes. Awesome job.

In addition to those comments, I found these bugs.

### sometimes I can't double click off when finishing drawing a way.
cannot-double-click-off

### I cannot drag nodes that belong to features within the currently active layer
cannot-drag-node

If you notice in the first gif how the icon changes. I think that indicates the mode is changing. For debugging why this is happening, it might be helpful to look at the context's `exit` and `enter` methods since they tell you when the modes are being switched.

jackgrossman18:
@maxgrossman @brianhatchl this is ready for review

Issue #1332, opened 2019-02-11
Replace external ref to google css with local resource


brianhatchl:
Will need to make this resource local for use on closed networks

PR #1436, opened 2019-05-22
all google fonts now local

jackgrossman18:
ref #1332


dist/index.html
```diff
@@ -4,8 +4,9 @@
         <meta charset='utf-8'>
         <title>Hootenanny</title>
         <link rel="icon" href="favicon.ico"/>
-        <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+        <link rel="stylesheet" href="../css/80_app.css">
```
maxgrossman: I think because all css referenced in the source code gets bundled into `iD.css`, this can be plucked. I just commented this out on my comp and seems to work fine.

PR #1438, opened 2019-05-22
Fixing ui test

bobi7:


modules/renderer/background_source.js
```diff
@@ -222,18 +222,18 @@ rendererBackgroundSource.Bing = function(data, dispatch) {
...
+        //         attribution: provider.attribution,
+        //         areas: provider.coverageAreas.map(function(area) {
+        //             return {
+        //                 zoom: [area.zoomMin, area.zoomMax],
+        //                 extent: geoExtent([area.bbox[1], area.bbox[0]], [area.bbox[3], area.bbox[2]])
+        //             };
+        //         })
+        //     };
+        // });
+        // dispatch.call('change');
```
brianhatchl: why this commenting out?

test/hoot/spec/datasetPanel/tableBehavior.js
```diff
@@ -36,49 +36,57 @@ module.exports = () => {
...
+                    parentId: 0,
+                    folderName: 'UnitTestFolder1'
+                };
 
-                });
+                await Hoot.api.addFolder( folderParams ); // generate test folder
+                const refresh = Hoot.folders.refreshAll();
 
+                return refresh;
+            } );
```
brianhatchl: So instead of a `before`, you move the dataset upload and folder add steps to their own test steps?

modules/renderer/background_source.js
bobi7: something was wrong with retrieving the link (checked on others machines as well) and it seemed to be causing test failures so i commented it out, but looks like it has been fixed so i readded this

test/hoot/spec/datasetPanel/tableBehavior.js
bobi7: i ran 2 tests at the end where 1 had it in the before and the latest didnt. they both seem to have passed. you think its better to keep it in before?

brianhatchl: i just want it to test the right behavior and pass consistently :grin:

bobi7: i think the only difference is if the before fails the stuff following it immediately fails too

PR #1439, opened 2019-05-24
Export UI updates

brianhatchl:


modules/Hoot/managers/api.js
```diff
@@ -701,19 +701,19 @@ export default class API {
...
+                // 'hoot:review:members':'',
+                // 'hoot:review:score':'',
+                // 'hoot:review:note':'',
+                // 'hoot:review:sort_order':'',
+                // 'hoot:review:type':'',
+                // 'hoot:review:needs':'',
+                // 'hoot:score:match':'',
+                // 'hoot:score:miss':'',
+                // 'hoot:score:review':'',
+                // 'hoot:score:uuid':''
```
brianhatchl: No need for default tag overrides removing hoot tags.  This is done by visitor params in the export command

modules/Hoot/managers/api.js
```diff
@@ -754,17 +753,15 @@ export default class API {
...
             } )
             .catch( (err) => {
                 console.log( err );
 
                 return {
-                    message: `Failed to export dataset: ${ data.input }`,
-                    status: 500,
-                    type: 'success'
+                    message: `Failed to export dataset: ${ data.outputname }`,
+                    type: 'error'
```
brianhatchl: These and similar all make the alert message for errors red instead of green.  Also the http status param was never used so removing

modules/Hoot/config/domMetadata.js
```diff
@@ -313,30 +332,28 @@ export function exportDataForm( zipOutput ) {
...
             class: 'hidden',
             inputType: 'checkbox',
-            value: 'append',
+            checked: false,
+            hidden: true
+        },
+        {
+            label: 'Include Hootenanny tags?',
+            id: exportHootTags,
+            inputType: 'checkbox',
```
brianhatchl: Makes the "Include Hootenanny tags" checkbox operable for OSM formatted exports.

css/hoot/modules/_forms.scss
```diff
@@ -87,6 +87,14 @@
...
         }
     }
+
+    h3 span {
+        width: 90%;
+        display: inline-block;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
```
brianhatchl: truncate export title with ellipsis

PR #1443, opened 2019-06-03
1442 . have /sizes/ endpoint send empty list to retrieve data for all layers

bobi7:
refs #1442


modules/Hoot/managers/api.js
```diff
@@ -510,11 +510,9 @@ export default class API {
     }
 
     getMapSizes( mapIds ) {
-        if ( !mapIds ) {
-            return null;
-        }
+        const ids = mapIds ? mapIds : '';
         const params = {
-            path: `/info/map/sizes?mapid=${ mapIds }`,
+            path: `/info/map/sizes?mapid=${ ids }`,
```
brianhatchl: Let's remove the `mapid` param all together

PR #1450, opened 2019-06-11
fix tools menu toggle

brianhatchl:


brianhatchl:
The menu slide transition is missing but we can revisit that later

modules/ui/tools.js
```diff
@@ -208,8 +208,10 @@ export function uiTools( context ) {
 
                 initDropdown();
             } else {
+                // console.log('toggle');
```
bobi7: remove

modules/ui/tools.js
```diff
@@ -219,32 +221,23 @@ export function uiTools( context ) {
...
-                //     bindBodyClick();
-                // } )
-                // ;
         }
 
-        function bindBodyClick() {
-            d3.select( 'body' ).on( 'click', () => toggle( () => initDropdown() ) );
+        function bindSingleBodyClick() {
+            d3.select( 'body' ).on( 'click', () => {
+                // console.log('body click');
```
bobi7: cleanup

PR #1460, opened 2019-06-13
Fixed #1459. added set height to review conflicts table and made it s…

### LINKED ISSUE #1459: make review feature table scrollable
ran into a case were there were a lot of tags so the table overflowed the screen

bobi7:
…crollable


brianhatchl:
I'm getting something weird in Chrome for review layers with few tags.
image

PR #1461, opened 2019-06-14
Allow for import/export of GeoPackage

jackgrossman18:
ref #1453

### DIFF
--- a/modules/Hoot/ui/modals/exportData.js
+++ b/modules/Hoot/ui/modals/exportData.js
@@ -98,7 +98,8 @@ export default class ExportData {
 
     getTranslationPath() {
         const selectedTranslation = this.translationSchemaCombo.node().value;
-        return this.translations.find( t => t.NAME === selectedTranslation ).PATH;
+        const translation = this.translations.find( t => t.NAME === selectedTranslation );
+        return !translation.hasOwnProperty('PATH')  ? translation.EXPORTPATH : translation.PATH;
     }
 
     getOutputType() {
@@ -106,7 +107,8 @@ export default class ExportData {
             'Shapefile': 'shp',
             'File Geodatabase': 'gdb',
             'OpenStreetMap (OSM)': 'osm',
-            'OpenStreetMap (PBF)': 'osm.pbf'
+            'OpenStreetMap (PBF)': 'osm.pbf',
+            'GeoPackage (GPKG)': 'gpkg'
         }[this.exportFormatCombo.node().value];
     }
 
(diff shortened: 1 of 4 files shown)


modules/Hoot/ui/modals/exportData.js
```diff
@@ -98,15 +98,17 @@ export default class ExportData {
 
     getTranslationPath() {
         const selectedTranslation = this.translationSchemaCombo.node().value;
-        return this.translations.find( t => t.NAME === selectedTranslation ).PATH;
+        const translation = this.translations.find( t => t.NAME === selectedTranslation );
+        return !translation.hasOwnProperty('PATH')  ? translation.EXPORTPATH : translation.PATH;
```
brianhatchl: Do you recall why this had to change?

jackgrossman18: the object OSM object w/in the object array that hoot-services returns for translations is different from the rest, it doesn't return a key 'PATH' with useful information, instead it returns 'EXPORTPATH'. I chose to make this change instead of altering the services code.

brianhatchl: What's the url that retrieves that?

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -57,6 +57,10 @@ export default class ImportDataset {
                 title: 'File (geonames, txt)',
                 value: 'GEONAMES'
             },
+            {
+                title: 'File (GeoPackage, gpkg)',
+                value: 'GPKG'
+            },
```
brianhatchl: I also wonder why we group `File (shp, zip, gdb.zip)` but we have separate entries for geojson, geonames, geopackage?

modules/Hoot/ui/modals/exportData.js
brianhatchl: Found it, http://localhost:8080/hoot-services/ingest/customscript/getlist

brianhatchl: I think this might fix a bug that we were actually relying on for correct behavior.  Meaning the services code expects OSM tag output export requests to have no `translation` param.  I'll need to investigate before merging this.

brianhatchl: Ok, made some changes in https://github.com/ngageoint/hootenanny/pull/3302 that allows for use of RenderDb.js translation for both osm and ogr output formats, so this fix should be ok.

PR #1467, opened 2019-06-20
show more detailed error message when importing/exporting

brianhatchl:
Old error popup:
Screenshot from 2019-06-20 14-39-58

New error popup:
Screenshot from 2019-06-20 14-40-33

### DIFF
--- a/modules/Hoot/ui/modals/exportData.js
+++ b/modules/Hoot/ui/modals/exportData.js
@@ -220,11 +220,19 @@ export default class ExportData {
             } )
             .catch( (err) => {
                 console.error(err);
+
                 let message = 'Error running export',
                     type = err.type,
                     keepOpen = true;
 
+                if (err.data.commandDetail.length > 0 && err.data.commandDetail[0].stderr !== '') {
+                    message = err.data.commandDetail[0].stderr;
+                }
+
                 Hoot.message.alert( { message, type, keepOpen } );
+            } )
+            .finally( () => {
+                Hoot.events.emit( 'modal-closed' );
             } );
         }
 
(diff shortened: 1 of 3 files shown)


modules/Hoot/ui/modals/exportData.js
```diff
@@ -220,11 +220,19 @@ export default class ExportData {
             } )
             .catch( (err) => {
                 console.error(err);
+
                 let message = 'Error running export',
                     type = err.type,
                     keepOpen = true;
 
+                if (err.data.commandDetail.length > 0 && err.data.commandDetail[0].stderr !== '') {
+                    message = err.data.commandDetail[0].stderr;
```
bobi7: Do we want to overwrite the initial error message and not append to it instead?

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -383,10 +383,15 @@ export default class ImportDataset {
             .then( () => Hoot.events.emit( 'render-dataset-table' ) )
             .catch( err => {
                 console.error(err);
+
                 let message = 'Error running import',
                     type = err.type,
                     keepOpen = true;
 
+                if (err.data.commandDetail.length > 0 && err.data.commandDetail[0].stderr !== '') {
+                    message = err.data.commandDetail[0].stderr;
```
bobi7: Same question as above

modules/Hoot/ui/modals/exportData.js
brianhatchl: The only reason I did that for import multiple was that the user would want to know which file upload caused the error.  For single import and export I thought the error message from the hoot command was sufficient.

bobi7: good point

Issue #1316, opened 2019-01-22
Import multiple files/directory to single map dataset


brianhatchl:
similar to https://github.com/ngageoint/hootenanny-ui/issues/923

Currently when importing multiple sources each gets written to it's own dataset.

PR #1470, opened 2019-06-24
added check in importMulti modal to add multi-file dataset as single …

jackgrossman18:
ref #1316

### DIFF
--- a/modules/Hoot/ui/modals/ImportMultiDatasets.js
+++ b/modules/Hoot/ui/modals/ImportMultiDatasets.js
@@ -118,6 +118,7 @@ export default class ImportMultiDatasets {
             this.updateButtonState();
         } else {
             this.asSingleLayerName.node().disabled = true;
+            this.updateButtonState();
         }
 
         // filter out geoname translations
--- a/modules/Hoot/ui/modals/ImportMultiDatasets.js
+++ b/modules/Hoot/ui/modals/ImportMultiDatasets.js
@@ -452,17 +452,17 @@ export default class ImportMultiDatasets {
         }
 
         this.processRequest = this.allProgress( proms, ( n, fileName ) => {
-                this.progressBar.property( 'value', n );
-                this.fileListInput
-                    .select( `option[value="${fileName}"]` )
-                    .classed( 'import-success', true );
-            } )
-            .then( () => Hoot.message.alert( {
-                message: 'All datasets successfully imported',
-                type: 'success'
-            } ) )
-            .then( () => Hoot.folders.refreshAll() )
-            .then( () => Hoot.events.emit( 'render-dataset-table' ) );
+            this.progressBar.property( 'value', n );
+            this.fileListInput
+                .select( `option[value="${fileName}"]` )
+                .classed( 'import-success', true );
+        } )
+        .then( () => Hoot.message.alert( {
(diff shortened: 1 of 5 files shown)


jackgrossman18:
@brianhatchl @maxgrossman this is ready for review

brianhatchl:
bruh :eyes: :fearful:

maxgrossman:
You can blame force pushes on me. also emoji less please.

I noticed that 2 of the previous hoot2x merges were brining in stale versions of the hoot2x branch (perhaps because jack's local hoot2x was out of sync, I'm not sure).

So I just suggested to roll back to the previous commit before those bad merges, cherry-pick the good commit after them and push to branch.

maxgrossman:
Though, I guess I should have made a point to do on a separate branch - that is my bad.

brianhatchl:
There was no :melvin: but I guess :neckbeard: is close enough.
Just wanted to make sure we almost never resort to using force push.
This workflow is important.
Whenever I merge I make sure to do a `git fetch` and then `git merge origin/<branch_name>` so it doesn't matter if my local branch has pulled.

jackgrossman18:
ultimately it's my fault, I'm working through getting the progress bar behavior you asked for
:hammer_and_wrench:

maxgrossman:
:muscle:

jackgrossman18:
@brianhatchl i updated the progress bar process so the success css behavior is the same for both uploading multiple layers and combining two layers as a single layer

modules/Hoot/ui/modals/ImportMultiDatasets.js
```diff
@@ -348,32 +365,30 @@ export default class ImportMultiDatasets {
...
-                    formData: this.getFormData( importFiles ),
+                    INPUT_NAME: asSingleName,
+                    formData: this.getFormData(this.fileIngest.node().files),
                     folderId
                 };
 
                 Hoot.api.uploadDataset( params )
                     .then( resp => Hoot.api.statusInterval( resp.data[ 0 ].jobid ) )
+                    .then( () => {
+                        this.progressBar.property('value', fileNames.length );
```
brianhatchl: just increment current value by 1 when a promise resolves.

modules/Hoot/ui/modals/ImportMultiDatasets.js
```diff
@@ -387,41 +402,73 @@ export default class ImportMultiDatasets {
...
+
+                    Hoot.api.uploadDataset( params )
+                        .then( resp => Hoot.api.statusInterval( resp.data[ 0 ].jobid ) )
+                        .then( () => {
+                            this.progressBar.property( 'value', index + 1 );
+                            this.fileListInput
+                                .select( `option[value="${name}"]` )
+                                .classed( 'import-success', true );
+                        } )
+                        .then( () => resolve( name ) )
```
brianhatchl: Not sure resolve is needed any longer as you've moved the callback code (updating progress bar and filename highlighting) directly into the promise.

However, I think something can be done to make this section of code more DRY.  The async method in both the asSingle and normal multiple upload are almost identical except for the filtering of files into each set to upload and the import-success selector.  I think we could just set a variable to skip the filtering of importFiles and the `[value=]' css selector

PR #1471, opened 2019-06-25
Zoom to review feature on tag-table click

jackgrossman18:
ref #1457

### DIFF
--- a/modules/Hoot/ui/conflicts/info.js
+++ b/modules/Hoot/ui/conflicts/info.js
@@ -135,7 +135,7 @@ export default class ConflictMetadata {
 
         if (panToId) {
             if ( feature.type === 'node' ) {
-                Hoot.context.map().centerZoom( extent.center(), 19);
+                 Hoot.context.map().centerZoom( extent.center(), 21);
             } else {
                 Hoot.context.map().centerZoom( extent.center(), Hoot.context.map().trimmedExtentZoom( extent ) - 0.5 );
             }
--- a/modules/Hoot/ui/conflicts/info.js
+++ b/modules/Hoot/ui/conflicts/info.js
@@ -11,6 +11,7 @@ import _forEach   from 'lodash-es/forEach';
 import _map       from 'lodash-es/map';
 import _startCase from 'lodash-es/startCase';
 import { modeSelect } from '../../../modes';
+import Map  from './map';
 
 export default class ConflictMetadata {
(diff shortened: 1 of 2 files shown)


modules/Hoot/ui/conflicts/map.js
```diff
@@ -43,10 +43,18 @@ export default class Map {
         // panning will cause a 'drawn' event to fire and will automatically highlight the nodes
         if ( panTo ) {
             this.panToConflict();
+
+            let feature1Click = d3.selectAll('td.value-col.feature1');
+            feature1Click.on('click', () => this.panToConflict());
+
+            let feature2Click = d3.selectAll('td.value-col.feature2');
+            feature2Click.on('click', () => this.panToConflict());
```
brianhatchl: I think a better home for this is in `info.js` where the tag table is created.  Also the behavior in Hoot1 includes selecting the entity.

See the code here:
https://github.com/ngageoint/hootenanny-ui/blob/develop/js/hoot/control/conflicts/info/ReviewTable.js#L63-L70

I also notice that after zooming, the extent snaps back to the entire extent of the review instead of just the individual feature.

jackgrossman18:
@brianhatchl I moved the 'onClick' event & recreated the panToEntity method within info.js

brianhatchl:
need to add the select entity code like in Hoot1

jackgrossman18:
@maxgrossman please check the zoom change when resolving individual reviews

modules/Hoot/ui/conflicts/info.js
```diff
@@ -84,18 +85,43 @@ export default class ConflictMetadata {
...
+        Hoot.context.enter(modeSelect(Hoot.context, [entity.id]));
+    }
+
+    /**
+     * Pan to feature with conflict
+     */
+
+    panToEntity(feature) {
+        let extent = feature.extent(Hoot.context.graph());
+        Hoot.context.map().centerZoom(extent.center(), 21);
```
maxgrossman: After merging hoot2x into this branch, there is [this](https://github.com/ngageoint/hootenanny-ui/blob/cf3e5b09633fd5cd698f21f11c059281008ffacf/modules/Hoot/ui/conflicts/map.js) function on the conflict's Map class.
We should use it here too.
However, I think we'll need to make it a static method so we can use it here without make a new instance of the class.

Issue #1456, opened 2019-06-13
complete review resolve all reviews does not clear review controls


bobi7:

jackgrossman18:
@bobi7 was this fixed?

bobi7:
@jackgrossman18  not that i know of

brianhatchl:
Turns out Complete Review... Resolve all remaining reviews - not working
The reviews relations are not removed because the popup prompting to go into review mode reappears.

PR #1473, opened 2019-06-26
adding 'resolveall' method to api

jackgrossman18:
ref #1456

### DIFF
--- a/modules/Hoot/managers/api.js
+++ b/modules/Hoot/managers/api.js
@@ -574,7 +574,7 @@ export default class API {
             .then( resp => resp.data );
     }
 
-    resolveAllConflicts(mapId) {
+    resolveAllReviews(mapId) {
         const params = {
             path: '/job/review/resolveall',
             method: 'PUT',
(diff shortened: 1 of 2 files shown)


jackgrossman18:
cc @maxgrossman

modules/Hoot/managers/api.js
```diff
@@ -574,6 +574,31 @@ export default class API {
...
+        const params = {
+            path: '/job/review/resolveall',
+            method: 'PUT',
+            data: {mapId: mapId}
+        };
+
+        return this.request( params )
+        .then( resp => {
+            if (resp.status !== 200) {
+                return Promise.reject({
```
maxgrossman: I think you can just do the alert stuff, no need for the internal promise.reject

modules/Hoot/ui/conflicts/resolve.js
```diff
@@ -85,7 +90,7 @@ export default class Resolve {
                 color: layer.color
             };
 
-        // enter controller refresh state
+        // enter controller refresh stats
```
maxgrossman: I think this was intended to say state.

jackgrossman18:
@maxgrossman made the changes you requested

PR #1482, opened 2019-07-08
update zooms for reviews so they are bound to level 16 and 21 and handle layer query param

### LINKED ISSUE #1479: Post-Conflation Does Not Kick Off Review Mode
**Behavior from smoke test**
- User initiates conflation
- Conflation completes, and prompts user into Review Mode
- User opts into review mode, but Hootenanny keeps the map zoomed out and does not initiate the Review Mode UI. 
- _I had to force a review mode by zooming into the geography I already knew data existed/where the first few data reviews would typically be._ 

**Expected behavior:**
- User initiates conflation
- Conflation completes, and prompts user into Review Mode
- User opts to initiate Review Mode, and Hootenanny zooms into the first data review (including the review mode UI controls)

maxgrossman:
closes #1479

### DIFF
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -12,7 +12,8 @@ import _forEach        from 'lodash-es/forEach';
 import _isEmpty        from 'lodash-es/isEmpty';
 import _map            from 'lodash-es/map';
 import _reduce         from 'lodash-es/reduce';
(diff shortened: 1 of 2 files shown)


maxgrossman:
This should make sure that very close features are not shown at zoom levels too high for background imagery to be meaningful, and for features very far apart from being shown at zoom level farther than what is allowed for rendering features.

PR #1483, opened 2019-07-08
do not allow undefined in hash, make hash match loadedLayers state

### LINKED ISSUE #1472: url shows undefined=*some id* when dataset added from datasets panel
I have seen this show up randomly but one way to reproduce this: go to datasets panel -> right click dataset and add it are secondary -> undefined will show up in url

maxgrossman:
closes #1472

### DIFF
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -200,9 +200,11 @@ export default class Layers {
             };
 
             //update url hash
-            var q = utilStringQs(window.location.hash.substring(1));
-            q[params.refType] = mapId;
-            window.location.replace('#' + utilQsString(q, true));
+            if (params.refType) {
+                var q = utilStringQs(window.location.hash.substring(1));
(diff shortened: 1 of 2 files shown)


modules/Hoot/managers/layerManager.js
```diff
@@ -287,16 +303,21 @@ export default class Layers {
                 input2Name = tags.input2Name;
 
             if ( (input1 && input1Name) && (input2 && input2Name) ) {
-                let params1 = {
-                    name: input1Name,
-                    id: input1,
-                    color: 'violet'
-                };
+                let loaded = Object.values(Hoot.layers.loadedLayers).find(l => l.id === tags.input1),
```
maxgrossman: This was where the undefined problem was occurring during the review,

maxgrossman: Since the params did not include a refType, then this gets passed to loadLayer, which tries to update hash using params.refType, but without checking if params.refType is undefined.

PR #1487, opened 2019-07-10
make dataset table bigger

### LINKED ISSUE #1484: Make dataset table bigger
The Dataset panel's fixed height can make looking for datasets difficult, especially if users add more and more layers. So, I think it'd be worthwhile to make the panel use the available window height.

maxgrossman:
closes #1484


maxgrossman:
### before 

image

### after 

image

Issue #1490, opened 2019-07-11
toggle merged with input layers not implemented in hoot2x


brianhatchl:
A conflated layer should be able to toggle to a "before" view of it's input layers.  Doesn't seem to have been implemented in Hoot2x

PR #1505, opened 2019-07-18
Before and after conflation toggle

maxgrossman:
This is still WIP, just want to see a diff

ref #1490


maxgrossman:
I think I got the toggle working, but it inconsistently will change which layer is rendered on top.
Will share pics.

maxgrossman:
Also, when I added the new button, the sidebar layer buttons' CSS were messed up.
I think this is due to a dynamic CSS width calculation that expected only 2 buttons, and not three.
So, I still need to take care of that, which I think we can try to accomplish with flex styling.
Ergo, this is still yet a WIP... if you will.

maxgrossman:
### can swap top layer

image

image


You might have to squint, but you can see the pink on top of the orange in the first, and the opposite in the second.

maxgrossman:
### cannot swap

image

image


Notice that while the toggle shows the one on top is different in the two photos, the rendering does not actually change

maxgrossman:
All that is to say, I think the score changes we made in #1414 need to be tweaked. That is, make the score boosts we give to the top layers bigger.

modules/svg/lines.js
```diff
@@ -94,18 +94,16 @@ export function svgLines(projection, context) {
 
 
         function waystack(a, b) {
+            // if top layer set, make top layer lines last to render.
+            var topLayer = Hoot.layers.getTopLayer();
+            if (topLayer) { return +a.mapId === topLayer ? 1 : -1 }
```
maxgrossman: when we are setting drawing order, use d3 / data array comparator function to make so the top layer map items are put at bottom of list and therefore rendered last.

modules/ui/background.js
```diff
@@ -337,7 +337,7 @@ export function uiBackground(context) {
                 d3.select(prev).select('button')
                     .classed('hidden', false);
 
-                Hoot.layers.setTopLayer(data.id);
+                Hoot.layers.setTopLayer(Number(data.id));
```
maxgrossman: make sure we are always comparing a number against a number

maxgrossman:
@brianhatchl - I think we are getting closer. I found what I think is a bug we want to avoid.
Each switch between before and after conflation layers resets the graph. So, any changes you make to one of the layers end up being lost if you switch between layers.

result-of-clearing-graph

My gif recorder stinks but "trust me" when I say I did just switch between the before/after layers.

I think this is where we need to use the history's 'toJSON' and 'fromJSON' so it will remember its states between toggles.

brianhatchl:
I made a few changes to include:

1. Allowing the merged view (with toggle) of the data even if there are no open reviews
2. Move the Hoot layer list controls from Background to Map Data
3. Update the icons in the layer entry to use material vs svg
4. Disable zooming to layer extent when toggling merged/input layers

If you want to have a play with this @maxgrossman

bobi7:
A bug I found reproducible with these steps

add a merged layer with reviews as reference
enter review move
open map data panel
turn on all layers visibility
you should see the merged layer should be last in this list
click the arrow icon for the merged layer to move it up

2 issues here. 
- 1 the arrow icon for the merge layer is gone even though its in the middle and not the top
- the reference layer is on the top but the merged layer is being shown instead

bobi7:
Run a conflate job that you know will result in reviews
after the conflate is done and the dialog box asks you if you want to enter review mode click cancel
the left panel will be cleared

brianhatchl:
In looking into your findings I also realize we can't allow Resolve a review to happen when input layers are loaded.

bobi7:
Looks good to me

brianhatchl (approved):

Issue #1509, opened 2019-07-22
Deleting several layers at once causes hoot-services error


jackgrossman18:
`/layers` in hoot services returns the following error

```
ERROR: relation "changesets_178" does not exist
```

brianhatchl:
#1511 doesn't fix this, but it's being tracked as a symptom of the tomcat deadlock ticket #1512 which we think is caused by query dsl misbehaving during concurrent requests that occur now that the UI uses Promise.all() and fires off requests (like upload and delete) all at once.

PR #1511, opened 2019-07-22
updating sorting of numbers that allows for #'d duplicate layers

jackgrossman18:
ref #1509


brianhatchl (approved):
Good catch!
I was not aware of this behavior https://alligator.io/js/array-sort-numbers/

Issue #1515, opened 2019-07-24
hide grail tools for non advanced users


bobi7:
companion ticket for: https://github.com/ngageoint/hootenanny/issues/3355

PR #1518, opened 2019-07-24
hide grail tools for non advanced users

bobi7:
#1515

### DIFF
--- a/modules/Hoot/ui/managePanel/jobs.js
+++ b/modules/Hoot/ui/managePanel/jobs.js
@@ -18,6 +18,8 @@ export default class Jobs extends Tab {
 
         this.name = 'Jobs';
         this.id   = 'util-jobs';
+
+        this.privileges = Hoot.user().privileges;
     }
 
     render() {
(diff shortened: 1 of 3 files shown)


brianhatchl:
```
1) Measurement and Clip tools appear

     UI tools

     Error: Uncaught AssertionError: expected 2 to equal 3 (webpack:///./node_modules/chai/lib/chai/assertion.js?:141)



2) All grail tools active

     UI tools

     Error: Uncaught AssertionError: expected 0 to equal 3 (webpack:///./node_modules/chai/lib/chai/assertion.js?:141)

```

We should give the Karma test user advanced privilege.

modules/ui/tools.js
```diff
@@ -15,6 +15,15 @@ import selectBbox       from '../Hoot/tools/selectBbox';
 import { tooltip }      from '../util/tooltip';
 
 export function uiTools( context ) {
+    let isAdvancedUser = false;
+    if ( localStorage ) {
+        const user = JSON.parse( localStorage.getItem( 'user' ) );
+
+        if (user && user.privileges) {
+            isAdvancedUser = user.privileges.advanced === 'true';
```
brianhatchl: I think the user object in local storage may encode the advanced value as a boolean.  We should double-check this.

modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -503,46 +505,50 @@ export default class Jobs extends Tab {
...
-                        }
-                    });
-                }
 
-                if (d.jobType.toUpperCase() === 'CONFLATE') {
-                    let currentLayer = this.findLayer( d.mapId );
-
-                    if (currentLayer && currentLayer.grail) {
+                // Only advanced user may perform these
+                if (this.privileges && this.privileges.advanced === 'true') {
```
brianhatchl: check if actual user object received from services and put in local storage encodes advanced as a boolean

brianhatchl: I think I answered my own question.  Because we are using hstore in the db, everything key and value are stored as string.

modules/ui/tools.js
brianhatchl: answered above

Issue #1516, opened 2019-07-24
Job Status polling interval should activate after successful response


brianhatchl:
Basically use `setTimeout` instead of `setInterval`.

Saw this when the UI kept issuing new requests when there were still existing pending requests when tomcat was frozen #1512.

Issue #1512, opened 2019-07-23
Cannot upload more than 5 files at once, even when size is less than ingestThreshold


jackgrossman18:
When attempting to upload more than 5 files at once, the importMulti modal becomes unresponsive and the files do not upload successfully. 

This requires a tomcat restart to resolve hanging upload.

brianhatchl:
See https://github.com/ngageoint/hootenanny/pull/3363

PR #1519, opened 2019-07-24
convert setInterval to setTimeout in job status polling

brianhatchl:
refs #1516


brianhatchl:
I decided to make this change while troubleshooting the tomcat freeze bug #1512.
When tomcat is frozen all the job status requests hang as pending.  Because `setInterval` was used the UI continues to poll the service even when there are outstanding pending requests.  Rather than continue this useless polling, I refactored to use `setTimeout` with a new polling request only being issued for the future when the last polling request returns a response.

Issue #1537, opened 2019-08-02
ImportMulit Modal  filtering files broken


jackgrossman18:
ImportMulti modal breaking when filtering files by name

brianhatchl:
Under what conditions do the names not match?

I'm concerned about file systems that support case sensitive names.

PR #1538, opened 2019-08-02
ImportMulti breaks when filtering file names

jackgrossman18:
ref #1537

### DIFF
--- a/modules/Hoot/ui/modals/ImportMultiDatasets.js
+++ b/modules/Hoot/ui/modals/ImportMultiDatasets.js
@@ -194,14 +194,7 @@ export default class ImportMultiDatasets {
 
     setFileMetadata( fileName, typeCount, fileList ) {
 
-        let fName = fileName.substring( 0, fileName.length - 4 );
-
-        if ( fileName.indexOf( '.shp.xml' ) > -1 ) {
-            fName = fileName.toLowerCase().substring( 0, fileName.length - 8 );
-        }
-        if ( fileName.indexOf( '.geojson' ) > -1 ) {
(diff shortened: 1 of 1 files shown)


modules/Hoot/ui/modals/ImportMultiDatasets.js
```diff
@@ -381,7 +381,7 @@ export default class ImportMultiDatasets {
                     //filter files by name if not merging to single layer
                     importFiles = _filter( files, file => {
 
-                        let fName = file.name.split('.')[0];
+                        let fName = file.name.toLowerCase().split('.')[0];
```
brianhatchl: I think this is a better fix to the issue:

```
@@ -192,18 +192,11 @@ export default class ImportMultiDatasets {
         this.updateButtonState();
     }
 
     setFileMetadata( fileName, typeCount, fileList ) {
 
-        let fName = fileName.substring( 0, fileName.length - 4 );
-
-        if ( fileName.indexOf( '.shp.xml' ) > -1 ) {
-            fName = fileName.toLowerCase().substring( 0, fileName.length - 8 );
-        }
-        if ( fileName.indexOf( '.geojson' ) > -1 ) {
-            fName = fileName.toLowerCase().substring( 0, fileName.length - 8 );
-        }
+        let fName = fileName.split('.')[0];
 
         let fObj = _find( fileList, file => file.name === fName );
 
         if ( !fObj ) {
             fObj = {
```
setFileMetaData was doing some sneaky stuff with geojson

brianhatchl (changes requested):

brianhatchl (approved):

Issue #1517, opened 2019-07-24
Clip path should inherit from layer being clipped


brianhatchl:
We are moving away from the default output path being `root`

PR #1547, opened 2019-08-12
default output folder now the same as layer that will be clipped

jackgrossman18:
ref #1517

### DIFF
--- a/modules/Hoot/tools/clipDataset.js
+++ b/modules/Hoot/tools/clipDataset.js
@@ -127,7 +127,7 @@ export default class ClipDataset {
                     } else if ( d.name === 'outputName' ) {
                         that.createLayerNameField( d3.select( this ), layer );
                     } else {
-                        var folderId = parseInt( Hoot.layers.findBy( 'id', layer.id ).folderId, 10 );
+                        var folderId = Hoot.layers.findBy( 'id', Number(layer.id) ).folderId;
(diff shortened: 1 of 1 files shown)


bobi7:
I see that the path placeholder for the clip output is the parent directory which is good but it should also be selected in the dropdown incase the user doesn't select it themselves.

jackgrossman18:
@bobi7

modules/Hoot/tools/clipDataset.js
```diff
@@ -128,7 +127,11 @@ export default class ClipDataset {
                     } else if ( d.name === 'outputName' ) {
                         that.createLayerNameField( d3.select( this ), layer );
                     } else {
+                        var folderId = Hoot.layers.findBy( 'id', Number(layer.id) ).folderId;
+                        d.combobox = [ d.combobox.find( function( l ) { return l.id === folderId; } ) ]
+                            .concat( d.combobox.filter( function( l ) { return l.id !== folderId; } ).sort() );
                         that.createFolderListCombo( d3.select( this ), d );
+                        d3.select( this ).property( 'value', Hoot.folders.findBy( 'id', folderId ).name );
```
bobi7: What happens if you also set the property '_value' to equal the result of folderId above? I feel like this may make is so that you dont need all the calls to the find() function below

jackgrossman18: :checkered_flag:

PR #1549, opened 2019-08-14
Fixed #1529. historical bbox added to grail pull modal

### LINKED ISSUE #1529: keep track of a history of bbox used for grail pull
when grail pull is selected we can have a dropdown that has the last couple used bbox's

bobi7:
ref #1529

### DIFF
--- a/modules/Hoot/tools/grailPull.js
+++ b/modules/Hoot/tools/grailPull.js
@@ -116,7 +116,10 @@ export default class GrailPull {
 
 
         let history = JSON.parse( Hoot.context.storage('history') );
-        history.bboxHistory.push( bbox );
+        if ( history.bboxHistory.length === 5 ) {
+            history.bboxHistory.pop();
+        }
+        history.bboxHistory.unshift( bbox );
         Hoot.context.storage( 'history', JSON.stringify( history ) );
 
         this.form.remove();
--- a/modules/Hoot/tools/grailPull.js
+++ b/modules/Hoot/tools/grailPull.js
@@ -187,8 +187,9 @@ export default class GrailPull {
 
 
         let history = JSON.parse( Hoot.context.storage('history') );
-        if ( history.bboxHistory.length === 5 ) {
-            history.bboxHistory.pop();
+        if ( history.bboxHistory.length >= 5 ) {
+            // Removes oldest (last in list) bbox
+            history.bboxHistory = history.bboxHistory.slice( 0, 4 );
         }
         history.bboxHistory.unshift( bbox );
         Hoot.context.storage( 'history', JSON.stringify( history ) );
--- a/modules/Hoot/tools/grailPull.js
+++ b/modules/Hoot/tools/grailPull.js
@@ -114,6 +114,11 @@ export default class GrailPull {
             .then( () => Hoot.folders.refreshAll() )
             .then( () => Hoot.events.emit( 'render-dataset-table' ) );
 
+
+        let history = JSON.parse( Hoot.context.storage('history') );
+        history.bboxHistory.push( bbox );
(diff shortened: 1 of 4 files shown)


bobi7:
Should i limit the amount of bboxes we keep in the history?

brianhatchl:
yah, 5?

modules/Hoot/tools/grailPull.js
```diff
@@ -114,6 +114,14 @@ export default class GrailPull {
             .then( () => Hoot.folders.refreshAll() )
             .then( () => Hoot.events.emit( 'render-dataset-table' ) );
 
+
+        let history = JSON.parse( Hoot.context.storage('history') );
+        if ( history.bboxHistory.length === 5 ) {
```
maxgrossman: I do not think anyone would have be clever and mutate the history to have > 5 entries, but they could. 
So should we do a greater than or equal to? 
And then instead of pop do a slice like `history.bboxHistory = history.bboxHistory.slice(1,5)`.

This ensures in the greater than or equal scenario we reset to first 4 bboxes before we add the new bbox on line 122

maxgrossman (changes requested):
Just the one comment. I also noticed a bug that's outside scope of this ticket but bbox related that I'll put in comments

maxgrossman:
If I select derive differential and click use visual extent I get this javascript error

```
selectBbox.js:216 Uncaught TypeError: Cannot read property 'node' of undefined
    at SelectBbox.handleBbox (selectBbox.js:216)
    at HTMLButtonElement.<anonymous> (selectBbox.js:152)
    at HTMLButtonElement.<anonymous> (on.js:27)
```

... the same happens if you draw bbox in derive changeset, maybe we just need to put that dropdown there in derive modal too?

modules/Hoot/tools/selectBbox.js
```diff
@@ -176,6 +212,10 @@ export default class SelectBbox extends EventEmitter {
         this.maxLonInput.property( 'value', this.maxlon );
         this.minLatInput.property( 'value', this.minlat );
 
+        if ( this.bboxSelectType !== 'boundsHistory' ){
+            this.dropdownContainer.node().value = '';
```
maxgrossman: Also, to wear my d3 badge on my chest, can we use some d3 here instead? 
```javascript
d3.select(...).attr('value', '')
```

maxgrossman:
Also, I created this error with these steps...

```
Uncaught TypeError: Cannot read property '0' of undefined
    at SelectBbox.updateCoords (selectBbox.js:60)
    at SelectBbox.handleBbox (selectBbox.js:207)
    at finish (clip_bounding_box.js:24)
    at Dispatch.call (dispatch.js:57)
    at ret (clip.js:118)
    at SVGSVGElement.<anonymous> (clip.js:34)
    at SVGSVGElement.<anonymous> (on.js:27)
```

1. select pull remote data
2. draw a bounds
3. click off the job
4. go back to pull remote data
5. select that bounds from my history
6. go to draw bounds
7. click my first corner

Also the mouse icon switches from the cross-hair to the regular mouse icon.

modules/Hoot/tools/grailPull.js
bobi7: true. better to be safe

modules/Hoot/tools/selectBbox.js
bobi7: 🤓 um actually...you would use `.property` instead but i know what you meant 🤓  
-@melvin

maxgrossman: lol

maxgrossman:
Big :+1: here - I think when you added dropdown to the differential changesets modal it made the undefined errors above no longer applicable

maxgrossman (approved):

PR #1554, opened 2019-08-15
Fixed #1553. set Hoot.layers.allLayers to empty array instead of object

### LINKED ISSUE #1553: job panel error when all datasets/folders are deleted
Clear all datasets and folders, go to jobs panel, and console will print a 'Hoot.layers.allLayers.find is not a function' error

bobi7:
ref #1553


brianhatchl:
Closing because could not reproduce on further inspection

Issue #1552, opened 2019-08-15
Differential road snap options missing from UI v2


bwitham:
The advanced options under "Differential Conflation Options" had some road snapping options, turned off by default in v1, that allowed for post conflate road snapping. They are missing in v2, so there is no way for users to activate that feature.

* differential.snap.unconnected.roads
* ~~differential.remove.reference.data~~
* snap.unconnected.ways.snap.tolerance
* snap.unconnected.ways.use.existing.way.nodes
* snap.unconnected.ways.existing.way.node.tolerance

bwitham:
@curranMapper

PR #1558, opened 2019-08-20
extend hide/show conflation-type options to include differential conf…

maxgrossman:
ref #1552

In the same way that we made attribute conflation enable/disable based on it being selected in the conflation form, we'll now do the same for differential

### DIFF
--- a/css/hoot/components/_advancedOptions.scss
+++ b/css/hoot/components/_advancedOptions.scss
@@ -49,19 +49,15 @@
     white-space: nowrap;
     transition: width 250ms ease-in-out, box-shadow 250ms ease-in, visibility 300ms;
 
-    & > div {
-        position: absolute;
-        width: 350px;
-    }
-
     &.visible {
         width: 350px;
         visibility: visible;
         transition: width 250ms ease-in-out;
     }
 
+
     .advanced-opts-header {
-        height: 60px;
+        height: 5%;
         padding: 15px;
         box-shadow: 0 1px 10px 0 $light2;
 
@@ -80,9 +76,9 @@
     }
 
     .advanced-opts-content {
-        top: 60px;
+        height: 95%;
         overflow-x: hidden;
-        overflow-y: auto;
+        overflow-y: scroll;
     }
 
     .advanced-opts-actions {
@@ -127,7 +123,7 @@
             .adv-opts-inner-wrap-left {
                 width: 20px;
                 align-items: center;
-                input { 
+                input {
                     margin-top: 5px;
                 }
             }
@@ -137,9 +133,7 @@
     .group-body {
         display: flex;
         flex-direction: column;
-        overflow: auto;
         align-items: center;
-        max-height: 500px;
         padding: 15px 10px;
     }
 
@@ -167,13 +161,13 @@
 
         input.combobox-input {
             display: inline-block;
-        }    
+        }
     }
 
     .hoot-form-field-checkbox {
         flex-direction: row;
         align-items: center;
-        
+
         .hoot-field-title-checkbox-wrap {
             display: flex;
             height: inherit;
@@ -183,8 +177,8 @@
             }
         }
         .hoot-field-input-checkbox-wrap {
-            width: 15%; 
-            height: inherit;    
+            width: 15%;
+            height: inherit;
             display: flex;
             justify-content: center;
             input {
@@ -246,6 +240,7 @@
 
 @media only screen and (max-width: 1160px) {
     #advanced-opts-panel {
+
         & > div {
             position: absolute;
             width: 300px;
--- a/modules/Hoot/config/domMetadata.js
+++ b/modules/Hoot/config/domMetadata.js
@@ -75,17 +75,19 @@ export function layerConflateForm( data ) {
                     advancedOpts.createGroups(advOpts);
                 } else {
                     // disable & enable the attribute conflation group.
-                    let attributeGroup = d3.select( '.advanced-opts-content #Attribute_group' ),
-                        isAttribute = d3.select( '#conflateType' ).property( 'value' ) === 'Attribute';
+                    [ 'Attribute', 'Differential' ].forEach((conflationGroup) => {
+                        let confGroup = d3.select( `.advanced-opts-content #${conflationGroup}_group` ),
+                            isGroup = d3.select( '#conflateType' ).property( 'value' ) === conflationGroup;
 
-                    attributeGroup.select( '.adv-opt-title' )
-                        .classed( 'adv-opt-title-disabled', !isAttribute );
+                        confGroup.select( '.adv-opt-title' )
+                            .classed( 'adv-opt-title-disabled', !isGroup );
 
-                    attributeGroup.select( '.adv-opt-toggle' )
-                        .classed( 'toggle-disabled', !isAttribute );
+                        confGroup.select( '.adv-opt-toggle' )
+                            .classed( 'toggle-disabled', !isGroup );
 
-                    attributeGroup
-                        .select( '.group-body', true );
+                        confGroup
+                            .select( '.group-body', true );
+                    })
                 }
             }
         },
--- a/modules/Hoot/config/domMetadata.js
+++ b/modules/Hoot/config/domMetadata.js
@@ -87,7 +87,7 @@ export function layerConflateForm( data ) {
 
                         confGroup
                             .select( '.group-body', true );
-                    })
+                    });
                 }
             }
         },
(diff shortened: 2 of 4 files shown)


brianhatchl:
we should make adv opts pane scrollable.
also getting a js error when attempting to click disabled accordion section (not caret):
```
Uncaught TypeError: Cannot read property 'checked' of null
    at Selection.property (property.js:27)
    at HTMLDivElement.<anonymous> (advancedOpts.js:186)
    at HTMLDivElement.<anonymous> (on.js:27)
```

maxgrossman:
Also, we should make the text and the button open the dropdown like you suggested! @brianhatchl

brianhatchl:
I kind of miss the text being able to uncheck (the text span doesn't fill up the entire accordion bar so should be able to let empty space expand the section).
image
Also, do we need the word `Options` everywhere?  If the checkboxes do indeed disable Buildings conflation, I think omitting the Options might make that more intuitive, though maybe we still need some other hint that the check disables certain conflation feature types.

And we should open a ticket for the build script that orders the conflation type (Attribute, Differential) specific options so they are all together.

maxgrossman:
I agree the script is a good idea because so that in the front end we can order the options...

Just a thought though that we are going to have to touch the service's config each time we want to update these options.

~Manually ordering the order of k/v pairs in the config in the same way we manually curate the options we want to have in the front end seems sufficient to me.~

I am convincing myself we should have this func. could be something like this

```javascript
['a','b','c','d','e'].reduce(function(lists,d) {
    if (['a','d'].includes(d)) {
    	lists[0].push(d)
    } else if (['b','e'].includes(d)) {
        lists[1].push(d)
    } else {
        lists[2].push(d)
    }

    return lists;
}, [[],[],[]]).flat()
```

where list[0] is general and cleaning, list[1] is conflation specific options, list[2] are all the feature-specific options

maxgrossman:
> I think omitting the Options might make that more intuitive, though maybe we still need some other hint that the check disables certain conflation feature types.

^ Maybe something like just putting features at the end? Though that might not be abundantly clear either, so perhaps it is time for an infobox or something that explains this.

css/hoot/components/_advancedOptions.scss
```diff
@@ -80,9 +76,9 @@
     }
 
     .advanced-opts-content {
-        top: 60px;
+        height: 95%;
```
maxgrossman, line 79: the overflow-y scroll lets us scroll just on the accordion section

modules/Hoot/managers/api.js
```diff
@@ -1260,9 +1260,7 @@ export default class API {
         const params = {
             url: `${ this.translationUrl }schema`,
             method: 'GET',
-            params: {
-                ...data
-            }
+            params: data
```
maxgrossman, line 1263: I kept getting lint errors, and it turns out that the `{...obj}` syntax is an es6 way to make a clone of an object.
I don't think we are doing anything except making a request with these param objects such that we are not creating vulnerability to the scenario where you create a bug because you update a property on an object you didn't know was used elsewhere.

So, I just removed the spread stuff... If we want to be careful I guess we could just do `Object.assign({},d)`.

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -161,10 +173,12 @@ export default class AdvancedOpts {
                 .attr( 'id', d => `${d.name}-toggle` )
                 .classed( 'conflate-type-toggle', true );
 
-
             innerInput.merge(innerInputEnter)
                 .property( 'checked', true )
-                .on( 'click', toggleOption );
+                .on('click', function(d) {
+                    let shouldHide = d3.select(this).property('checked');
+                    instance.toggleOption(d, shouldHide);
```
maxgrossman, line 178: The toggleOption function is fired by clicking the label text too. 

So, the func has a weird quality of needing to update the state of the checkbox and so here, when you click a checkbox we don't just rely on vaniilla web browser to decide if the checkbox is checkked or not. We check the checkbox state ourselves and drive it property that way.

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -202,7 +206,8 @@ export default class AdvancedOpts {
         innerLabel.merge(innerLabelEnter)
             .attr( 'id', d => `${ d.name }_label` )
             .classed( 'adv-opt-title-disabled', false )
-            .text( d => `${d.label} Options` );
+            .classed( 'adv-opts-group-title', true)
+            .text( d => d.members.length ? `${d.label} Options` : d.label);
```
maxgrossman, line 208: if accordion does not have any advanced options we just don't add the options label

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -218,18 +223,47 @@ export default class AdvancedOpts {
...
-            });
+            .classed( 'combobox-caret', d => d.members.length );
 
         caretWrap.merge(caretWrapEnter);
+        caretWrap.on( 'click', instance.showBody);
+    }
+
+    showBody(d) {
+
+        if (d3.event.target.classList.contains( 'conflate-type-toggle' )) return;
```
maxgrossman: Ignore the showBody function from firing if we click the checkbox.
Otherwise, the toggleOption function gets called twice and the user just experiences their click to be a noop.

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -218,18 +223,47 @@ export default class AdvancedOpts {
...
+
+
+        if (d3.event.target.classList.contains( 'adv-opts-group-title' )) {
```
maxgrossman: how we let clicking the accordion text fire the code to disable the clicked accordion

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -218,18 +223,47 @@ export default class AdvancedOpts {
...
+            let input = d3.select( `#${d.name}-toggle` ),
+                shouldHide = !input.empty() && !input.property('checked');
+            instance.toggleOption(d, shouldHide, true);
+        } else if (d.members.length) {
+            let bodyState = d3.select( `#${ d.name }_group .group-body` ).classed( 'hidden' );
```
maxgrossman, line 246: If you are clicking anywhere else or on the combobox caret, then we update classes on all the accordions such that only the accordion you remains open and the rest get closed

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -218,18 +223,47 @@ export default class AdvancedOpts {
...
+            d3.selectAll('.advanced-opts-content .form-group .group-body')
+                .classed('hidden', function(data) {
+                    if (data.name === d.name) {
+                        let disabled = d3.select(this.parentElement)
```
maxgrossman, line 250: I am not sure if others have feelings on this, but this makes the behavior that if you have an accordion open, clicking that accordion's checkbox or label acts as a noop. you can only close it and then disable it (why you might click the checkbox) if you either click on its caret/whitespace or open up another accordion.

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -351,7 +385,7 @@ export default class AdvancedOpts {
                         d.send = value !== d.default;
                         if ([ 'double', 'int', 'long' ].indexOf ( d.type ) !== -1 ) {
                             d3.select( `#${d.id}-label-wrao` )
-                                .call(self.notNumber, value);
+                                .call(instance.notNumber, value);
```
maxgrossman, line 392: all these places I use instance was just a way to stop passing `this/that/self` objects around since we set the equivalent to the `instance` variable when you init the class.

modules/Hoot/ui/sidebar/advancedOpts.js
bobi7: is there a reason we dont do something 
- user clicks check box while accordion is open
- trigger a close accordion
- after close unselect it

modules/Hoot/config/domMetadata.js
```diff
@@ -75,17 +75,19 @@ export function layerConflateForm( data ) {
                     advancedOpts.createGroups(advOpts);
                 } else {
                     // disable & enable the attribute conflation group.
-                    let attributeGroup = d3.select( '.advanced-opts-content #Attribute_group' ),
-                        isAttribute = d3.select( '#conflateType' ).property( 'value' ) === 'Attribute';
+                    [ 'Attribute', 'Differential' ].forEach((conflationGroup) => {
+                        let confGroup = d3.select( `.advanced-opts-content #${conflationGroup}_group` ),
+                            isGroup = d3.select( '#conflateType' ).property( 'value' ) === conflationGroup;
```
bobi7: do something like `d3.select( '#conflateType' ).property( 'value' ).includes(conflationGroup)` so that Differential w/ tags will also cause the differential group to get enabled

bobi7 (changes requested):

bobi7:
looks good 🎸

bobi7 (approved):

PR #1561, opened 2019-08-21
show export and add new layer buttons for no-conflict merged layers

### LINKED ISSUE #1559: expose same export options we do in complete reivew when new reviews exist
There are some nice functions (like export data or add another dataset).
We should provide these functionalities always!

maxgrossman:
closes #1559


bobi7 (approved):
looks good 🚀

Issue #1563, opened 2019-08-21
add cancel jobs to import multiple


bobi7:
the cancel would only cancel the currently running jobs and do a cleanup for those while leaving the completed untouched

PR #1565, opened 2019-08-22
Fixed #1564 Clip job cancel for multiple jobs fixed

### LINKED ISSUE #1564: data cleanup if error occurs
ui companion ticket for https://github.com/ngageoint/hootenanny/issues/3270

bobi7:
ref #1564


modules/Hoot/ui/modals/importDataset.js
```diff
@@ -369,11 +369,17 @@ export default class ImportDataset {
...
                 let message;
                 if (resp.data && resp.data.status === 'cancelled') {
                     message = 'Import job cancelled';
 
+                    // Delete the newly created folder
+                    if ( newFolderName ) {
+                        await Hoot.api.deleteFolder( folderId );
+                        await Hoot.folders.removeFolder( folderId );
+                    }
+
```
brianhatchl, line 382: We talked about how this would be better done within the services.  If we moved folder creation to within a job command we could have that occur after successful clipping and if it was canceled the new output folder would never be created so we would not have to clean it up at all.  However, the services don't have a mechanism for passing input between job commands, so the folder id created by the AddFolder command would not be available to the UpdateParent command that moves output maps underneath said folder.
So we'll keep this way for now as it should run immediately after canceling.

brianhatchl (approved):
Looks good, I actually thought this was adding Cancel for multi-import, but I can do that next and will create a ticket.

brianhatchl:
I see it's already there #1563

modules/Hoot/ui/modals/importDataset.js
bobi7: Note: moved this to https://github.com/ngageoint/hootenanny/issues/3412

PR #1568, opened 2019-08-23
group & sort advanced options list elements.

### LINKED ISSUE #1567: ensure adv options order with sort function
It'd be helpful to have control of advanced options list sort order. we should add some code to the front end to do this.

maxgrossman:
closes #1567


brianhatchl (commented):
what if we gave these an order property in the json and just used that to sort?

maxgrossman:
Maybe, but if we imply sort by the order of the objects in that file can't we just do that.

maxgrossman:
rather than add more code.

brianhatchl:
Even better, let's just reorder https://github.com/ngageoint/hootenanny/blob/master/conf/services/conflationTypes.json

PR #1576, opened 2019-08-29
Show review mode for merged layer without inputs metadata

### LINKED ISSUE #1575: Layers with reviews but without map tags including input1 and input2 should still offer review mode
when exporting and importing a merged layer between two hoot instances, the map tags including metadata like input1, input2, params, stats, etc will not transfer, but review mode should still be available.

brianhatchl:
closes #1575


bobi7 (approved):
🚀

PR #1577, opened 2019-09-01
Fixed #1574. show folder size warning on export

### LINKED ISSUE #1574: on large folder export show warning

bobi7:
ref #1574


bobi7:
Should I just add the max export size in the services code?

modules/Hoot/ui/modals/exportData.js
```diff
@@ -54,6 +66,31 @@ export default class ExportData {
...
+     * @returns total folder size in bytes
+     */
+    calculateFolderSize ( root ) {
+        let stack = [ root ];
+        let totalSize = 0;
+
+        while ( stack.length > 0 ) {
+            const folder = stack.pop();
+            // children are stored in different locations in the object based on whether the folder is open or not
+            const children = folder.children || folder._children || folder.data._children;
```
brianhatchl: This feels kinda gross.  Maybe there's a better way via the d3 data binding methods on the foldertree object (rather than spelunking the internals).  I also got an error when an empty folder was included in the export and inspecting the folder object led me to see that there is a circular reference with the data property and itself (folder) which we should clean up.

modules/Hoot/ui/modals/exportData.js
```diff
@@ -44,6 +48,14 @@ export default class ExportData {
 
         if ( this.type === 'Datasets' ) {
             this.dataExportNameTextInput.attr( 'placeholder', this.input.split(',').join('_') );
+        } else if ( this.type === 'Folder' ) {
+            const folderSize = this.calculateFolderSize( this.data );
+
+            if ( folderSize > this.maxExportSize ) {
+                this.container.select( 'form' ).append( 'div' ).text(
+                    `Max export size of ${ this.maxExportSize } exceeded with current export size of ${ folderSize }`
```
brianhatchl: I think we want to call `formatSize()` on these values.

modules/Hoot/ui/modals/exportData.js
```diff
@@ -18,6 +19,9 @@ export default class ExportData {
         this.id = isDatasets ? d.map(n => n.id).join(',') : d.data.id;
         this.type = type;
         this.form = exportDataForm.call(this, isDatasets );
+        this.data = d;
+
+        this.maxExportSize = 500000000; // 500 Megabytes
```
brianhatchl: Maybe 2G limit?  If you just pulled that number out I'd vote to bump it up.

brianhatchl (approved):

PR #1580, opened 2019-09-03
Fixed #1579. fixed checkbox bug that caused map data toggle option to…

### LINKED ISSUE #1579: Toggling off conflated input layer, closing Map Data, causes checkbox to disappear
Then toggling conflated and input layers, a set of layer entries appears in the Map Data pane.  Toggling off one of the input layer entries and then closing the pane causes it to be removed and not restored when the Map Data pane is reopened.

bobi7:
… dissapear

ref #1579

### DIFF
--- a/css/25_areas.css
+++ b/css/25_areas.css
@@ -2,7 +2,7 @@
 /* Default - light gray */
 path.area.stroke {
     stroke: #ddd;
-    stroke-width: 1;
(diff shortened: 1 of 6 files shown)


brianhatchl (approved):
:hot_pepper: :hot_pepper: :hot_pepper:

PR #1585, opened 2019-09-10
look through all advanced options groupings to set defaults based on …

### LINKED ISSUE #1583: Advanced options show incorrect default values when Attribute Conflation is selected
Attribute Conflation defaults `building.keep.more.complex.geometry.when.auto.merging=false` as defined in AttributeConflation.conf. Select "Attribute Conflation" as the conflation type from the UI, open the Advanced Options and you see "Keep more complex geometry when merging" set to true under "Building Options".

maxgrossman:
…configs

closes #1583


bobi7 (approved):
LGTM

PR #1588, opened 2019-09-12
ctrlKey event for selecting/deselecting files in folder tree

jackgrossman18:
ref #1582


brianhatchl (approved):
Seems like the ctrl+click code was originally in the wrong spot.  The `bindContextMenu` function is only used by right-click, so the `d3.event.which === 1` (left-click) would never be reached.
I also had to remove datasets from the selection when deselected with a ctrl+click.

Issue #1578, opened 2019-09-03
Add a method to visualize a hoot changeset on the map


brianhatchl:
I think this would involve adding the reference layer to the map and highlighting the deleted and modified features and adding the merged layer to the map and highlighting the created and modified features

PR #1594, opened 2019-09-23
creating class for sorting modify, create and deletes

jackgrossman18:
ref #1578


brianhatchl:
Closing until we revisit

PR #1596, opened 2019-09-25
fixed #1589. allow user to sent custom query for overpass pull and derive

### LINKED ISSUE #1589: add text box so user can use their custom overpass query
add wherever overpass would be used ie. grail pull, derive diff changeset

bobi7:
ref #1589


modules/Hoot/managers/api.js
```diff
@@ -1111,8 +1111,9 @@ export default class API {
 
     grailPullOverpassToDb( data ) {
         const params = {
-            path: `/grail/pulloverpasstodb?bbox=${ data.BBOX }&name=${ data.secondaryName }`,
-            method: 'GET'
+            path: '/grail/pulloverpasstodb',
+            method: 'POST',
```
bobi7, line 1089: changed some of these to post so we could send the customQuery through the data instead of in the url because smaller size limit there

modules/Hoot/tools/selectBbox.js
```diff
@@ -200,6 +200,34 @@ export default class SelectBbox extends EventEmitter {
...
+                .classed( 'button-wrap user-input', true );
+
+            this.userInputContainer.append( 'input' )
+                .attr( 'type', 'checkbox' )
+                .on('click', function() {
+                    const isChecked = d3.select( this ).property( 'checked' );
+                    customQueryInput.classed( 'hidden', !isChecked );
+                });
+
+            this.userInputContainer.append('div').text('Use custom query for public overpass data');
```
brianhatchl: Let's change the text to just "Overpass query"

modules/Hoot/tools/selectBbox.js
```diff
@@ -200,6 +200,34 @@ export default class SelectBbox extends EventEmitter {
...
+
+            const placeholder = 'Example Query:\n' +
```
brianhatchl: Remove "Example Query:" line

brianhatchl (changes requested):
I think we need a "Back" button in case our overpass query returns no features.

modules/Hoot/tools/selectBbox.js
```diff
@@ -200,6 +200,34 @@ export default class SelectBbox extends EventEmitter {
...
+                '[out:json][bbox:{{bbox}}];\n' +
+                '(\n' +
+                '   node;<;>;\n' +
+                ')->.all;\n' +
+                '.all;\n' +
+                'out meta;\n' +
+                '>;\n';
```
brianhatchl: This is what Overpass Turbo uses for simple bbox call
```
/*
This is a simple map call.
It returns all data in the bounding box.
*/
[out:xml];
(
  node({{bbox}});
  <;
);
out meta;
```

what's the diff?

brianhatchl (changes requested):
Let's talk about moving the overpass query to a second window that appears after Next and before the count query runs.
We also may want to move the bbox extent options to a list box instead of buttons to accommodate "Use Reference" and "Use Secondary".
We may also want a validation to check for `[out:xml];` and replace with `[out:json];` and also check for the presence of `out meta;`

modules/Hoot/tools/selectBbox.js
bobi7: it doesnt recurse back down so you can potentially lose out on relations that may not have nodes.

the bbox declaration at the top is nice because you dont need to keep stating `({{bbox}})`

I will clean the example up a little though

bobi7:
I will add the stats for the private overpass pull results as well to this PR

modules/Hoot/tools/selectBbox.js
```diff
@@ -15,17 +15,17 @@ import { d3combobox }             from '../../lib/hoot/d3.combobox';
...
-        this.maxlat         = null;
-        this.bboxSelectType = 'visualExtent';
+        this.context        = predefinedData && predefinedData.context ? predefinedData.context : context;
+        this.minlon         = predefinedData && predefinedData.minlon ? parseFloat(predefinedData.minlon) : null;
+        this.minlat         = predefinedData && predefinedData.minlat ? parseFloat(predefinedData.minlat) : null;
+        this.maxlon         = predefinedData && predefinedData.maxlon ? parseFloat(predefinedData.maxlon) : null;
+        this.maxlat         = predefinedData && predefinedData.maxlat ? parseFloat(predefinedData.maxlat) : null;
+        this.bboxSelectType = predefinedData && predefinedData.bboxSelectType ? predefinedData.bboxSelectType : 'visualExtent';
         this.operationName  = '';
+        this.selectedBoundOption = predefinedData && predefinedData.selectedBoundOption ? predefinedData.selectedBoundOption : 'Visual Extent';
```
bobi7, line 28: this stuff is pretty ugly but couldnt figure out a really nice way. The other option was to just set the defaults here and maybe in a helper function replace the defaults with the 'predefinedData' if it was defined

bobi7:
Tasks to be done:

- [x] filter for private overpass
- [x] success message should use data source labels from overpass meta request
- [x] if map empty add pulled layers to ref/sec
- [x] move spinner from submit button to center of modal
- [x] leave modal open on grail pull submit. close when both jobs done
- [x] custom query checkbox label should also be checkable

brianhatchl (approved):

PR #1605, opened 2019-10-08
Add export dataset option for alpha shape and task grid geojson

### LINKED ISSUE #1598: Resurrect creating conflation campaign from dataset panel
refs hoot1 implementation #744 

make it available for advanced users only in hoot2x

brianhatchl:
closes #1598 

Instead of trying to POST a project json directly to Tasking Manager, I've chosen to add two export options, one for alpha shape of the "to-be" secondary dataset that can be used as the AOI for a new TM project.  Another for node density balanced rectangles that can optionally be clipped to the alhpa shape and be used as the arbitrary geometry grids in a new TM project.

At the same time as these geojson files are exported, they can also be added to the map so the user can make sure the generated output is suitable before importing into TM.  They may choose to re-run the generated shapes with different input parameters.


modules/Hoot/ui/modals/exportData.js
```diff
@@ -222,8 +222,8 @@ export default class ExportData {
                 includehoottags: self.includeHootTagsCheckbox.property( 'checked' ),
                 outputname: self.getOutputName(),
                 outputtype: self.getOutputType(),
-                tagoverrides: {},
-                textstatus: false,
+                // tagoverrides: "{}",
+                // textstatus: false,
```
bobi7, line 228: remove

bobi7 (changes requested):
just my 1 nitpick

modules/Hoot/ui/modals/exportData.js
brianhatchl: I'm gonna leave those, but will add a comment that they are unimplemented, i.e. they are hardcoded props for features from hoot1 that didn't make it into hoot2x yet.

PR #1608, opened 2019-10-09
remove reliance on main hoot repo for jenkins tests

brianhatchl:
refs #1607


brianhatchl:
@sputnam So this PR did end up having to copy some dependencies from the main hoot repo, but might be worth the change just so the git commit messages actually reflect the hoot-ui branch under test.

brianhatchl:
Ok, so this just required a Vagrantfile and a change in test expectations for the test user having advanced privilege.

sputnam (approved):
I like it. Looks good.

Issue #1611, opened 2019-10-16
login.html redirect should not lose querystring params


brianhatchl:
when opening hoot ui from tasking manager, and an active hoot session does not exist the uses will be redirected to login.html.  but when returning to index.html the querystring parameters like gpx get lost.

PR #1616, opened 2019-10-22
restore url hash from localStorage

jackgrossman18:
ref #1611

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -25,6 +25,10 @@
 
             Hoot.api.getServicesVersionInfo()
                 .then( () => {
+                    // apply cached hash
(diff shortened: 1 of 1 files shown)


dist/index.html
```diff
@@ -25,6 +25,10 @@
 
             Hoot.api.getServicesVersionInfo()
                 .then( () => {
+                    // apply cached hash
+                    window.location.hash = localStorage.getItem('hootQueryString');
```
brianhatchl: We don't always want to do this because if there's not a query string in local storage (such as when already logged in) this will whack the existing hash that has things like map location/zoom.

brianhatchl (changes requested):

brianhatchl (approved):

Issue #1614, opened 2019-10-17
Multi-select for folders in datasets panel would be nice


brianhatchl:

jackgrossman18:
This ticket allows for multi-select of folders in the dataset modal. 

This new feature required several updates to the datasets, modifyFolder and folderTree classes. 

**Dataset**

Adding multi-select folder feature to the folderContextMenu required updating this class so that it can pass all the selected folders to the modifyFolder class, previously, it would just pass the single folder to the modifyFolder class, now it passes the selected folders as an object array.

**modifyFolders**

Selecting multiple folders requires this class to filter through all of the current Hoot folders, then return only those folders that make sense as a destination, this is a list of the prevented destinations.

_Prevented Destinations_

1. Either of the selected folders.
2. Any of the children of the selected folders.
3. The parent of either of the selected folders 

_Acceptable Destinations_
1. Root
2. Sibling folder of the selected folders.
3. Any folder that is a sibling of the parent of the selected folder(s)

**folderTree**

The modify folder class required a few updates to handle the application of the new ctrl key event, previously only used to select multiple datasets, now listening for the selection of multiple folders.

The changes made also required updates to assure that they were congruous with both the new multi-select dataset action, and the additions recently made to the folderTree ( #1626 & #1627). 

When multiple folders are selected, the `folderContextMenu` only allows for thee options. This is because the list of options that are available when a single folder is selected contains a few that aren't relevant. 

_Available options_

1. Delete the selected folders
2. Move the selected folders
3. Export the selected folders

_Removed options_

1. The user shouldn't be able to rename the folders that they are moving, this is better suited when only a single folder is selected.
2. Adding either a single or multiple datasets makes more sense, and avoids problems, when a single folder is selected.
3. Adding a folder to two separate folders doesn't seem like the best idea.

PR #1620, opened 2019-10-25
updating css for selected folders, delete folders and their datasets

jackgrossman18:
ref #1614

### DIFF
--- a/modules/Hoot/tools/folderTree.js
+++ b/modules/Hoot/tools/folderTree.js
@@ -548,7 +548,7 @@ export default class FolderTree extends EventEmitter {
         let { data } = d,
             opts;
 
-        if ( data.type === 'dataset' || data.type === 'folder' && data.selected ) {
+        if ( data.type === 'dataset' ) {
             const selectedCount = this.selectedNodes.length;
 
             opts = [
@@ -595,6 +595,14 @@ export default class FolderTree extends EventEmitter {
                 }
             }
         } else if ( data.type === 'folder' ) {
+            if ( data.selected ) {
+                opts = [ ...this.folderContextMenu.slice() ]; // make copy of array to not overwrite default vals
+                opts.splice( 1, 0, {
+                    title: `Modify Folder ${ data.name }`,
+                    _icon: 'info',
+                    click: 'modifyFolder'
+                } );
+            }
             opts = [ ...this.folderContextMenu.slice() ]; // make copy of array to not overwrite default vals
             opts.splice( 1, 0, {
                 title: `Modify Folder ${ data.name }`,
--- a/modules/Hoot/tools/folderTree.js
+++ b/modules/Hoot/tools/folderTree.js
@@ -541,7 +541,7 @@ export default class FolderTree extends EventEmitter {
 
     checkMultiSelectFolder( d ) {
         if ( d.type === 'folder' && d.selected && this.selectedNodes.length > 1 ) {
-            return d.name + ' ...';
+            return d.name + '...';
         }
     }
 
@@ -603,7 +603,7 @@ export default class FolderTree extends EventEmitter {
         } else if ( data.type === 'folder' ) {
                 opts = [ ...this.folderContextMenu.slice() ]; // make copy of array to not overwrite default vals
                 opts.splice( 1, 0, {
-                    title: `Modify Folder ${ this.selectedNodes.length > 1 ? this.checkMultiSelectFolder(data) : data.name }`,
+                    title: this.selectedNodes.length ? `Move Folders ${ this.checkMultiSelectFolder(data) }` : `Modify Folder ${ data.name }`,
                     _icon: 'info',
                     click: 'modifyFolder'
                 } );
--- a/modules/Hoot/tools/folderTree.js
+++ b/modules/Hoot/tools/folderTree.js
@@ -587,7 +587,7 @@ export default class FolderTree extends EventEmitter {
             if ( selectedCount === 1 ) {
                 opts = [ ...this.folderContextMenu.slice() ]; // make copy of array to not overwrite default vals
                 opts.splice( 1, 0, {
-                title: `Modify/Move Folder`,
+                title: 'Modify/Move Folder',
                 _icon: 'info',
                 click: 'modifyFolder'
                 } );
--- a/modules/Hoot/tools/folderTree.js
+++ b/modules/Hoot/tools/folderTree.js
@@ -688,9 +688,7 @@ export default class FolderTree extends EventEmitter {
                 this.lastBasePosition = basePosition;
             }
             else if ( d3.event.ctrlKey && this.isDatasetTable ) {
-
                 data.selected = !data.selected;
-
                 if (data.selected) {
                     this.selectedNodes.push( data );
                 } else {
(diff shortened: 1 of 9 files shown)


jackgrossman18:
@maxgrossman

modules/Hoot/tools/folderTree.js
```diff
@@ -436,7 +436,12 @@ export default class FolderTree extends EventEmitter {
         let { data } = d;
 
         if ( data.type === 'folder' ) {
-            return (data.public) ? '#7092ff' : '#efefef';
+            if ( data.selected ) {
+                return '#ffff99';
+            }
+            else {
```
brianhatchl: don't really need the else here since the if has a `return`.

modules/Hoot/tools/folderTree.js
```diff
@@ -534,6 +539,12 @@ export default class FolderTree extends EventEmitter {
         this.update( d );
     }
 
+    checkMultiSelectFolder( d ) {
```
brianhatchl: should this be called `ellipsizeFolderNames`?
what should it return if not more than one selected?  or it's not selected?  or it's type is not folder?

modules/Hoot/tools/folderTree.js
```diff
@@ -590,12 +601,12 @@ export default class FolderTree extends EventEmitter {
...
-                title: `Modify Folder ${ data.name }`,
-                _icon: 'info',
-                click: 'modifyFolder'
-            } );
+                opts = [ ...this.folderContextMenu.slice() ]; // make copy of array to not overwrite default vals
+                opts.splice( 1, 0, {
+                    title: this.selectedNodes.length ? `Move Folders ${ this.checkMultiSelectFolder(data) }` : `Modify Folder ${ data.name }`,
+                    _icon: 'info',
+                    click: 'modifyFolder'
+                } );
```
brianhatchl: No need for these indents.
What's the new behavior here?
Add a comment if multiple folder selections can't be renamed whilst modifying.

modules/Hoot/tools/folderTree.js
```diff
@@ -643,7 +654,7 @@ export default class FolderTree extends EventEmitter {
             selected = data.selected || false,
             isOpen   = data.state === 'open';
 
-        if ( data.type === 'dataset' ) {
+        if ( data.type === 'dataset' || data.type === 'folder' && d3.event.ctrlKey ) {
```
brianhatchl: are there just two `data.type` values?  folder and dataset?
is there still a need for this || conditional?

modules/Hoot/tools/folderTree.js
```diff
@@ -727,7 +737,8 @@ export default class FolderTree extends EventEmitter {
                     d.children     = null;
                     data.selected  = false;
                 }
-            } else {
+            }
+            else {
```
brianhatchl: nitpick: I don't really care for this syntax using two lines

modules/Hoot/ui/managePanel/datasets.js
```diff
@@ -247,7 +247,7 @@ export default class Datasets extends Tab {
 
                 if ( !confirm ) return;
 
-                let items = d.data.type === 'folder' ? new Array( d ) : tree.selectedNodes;
+                let items = d.data.type === 'folder' && !d.data.selected ? new Array( d ) : tree.selectedNodes;
```
brianhatchl: let's put some parentheses here to capture the conditional part of the ternary.

modules/Hoot/ui/managePanel/datasets.js
```diff
@@ -310,10 +310,16 @@ export default class Datasets extends Tab {
                 break;
             }
             case 'modifyFolder': {
-                this.modifyFolderModal = new ModifyFolder( d ).render();
 
+                if ( tree.selectedNodes.length > 1 ) {
+                   this.modifyFolderModal = new ModifyFolder( tree.selectedNodes ).render();
+                }
+                else {
```
brianhatchl: syntax fix

modules/Hoot/ui/managePanel/datasets.js
```diff
@@ -310,10 +310,16 @@ export default class Datasets extends Tab {
...
+                    this.modifyFolderModal = new ModifyFolder( [d] ).render();
```
brianhatchl: would `tree.selectedNodes` not return a single item array here?

modules/Hoot/ui/modals/modifyFolder.js
```diff
@@ -12,40 +12,55 @@ import { modifyDatasetForm } from '../../config/domMetadata';
...
+
+            this.form.splice( 0, 1 );
+
+            let that = this;
+
+            that = that.folderList.filter( f => this.data.findIndex( folder => folder.parentId === f.id || folder.id === f.id || folder.name === f.name ) === -1 );
+
+            this.form[0].data = that;
+
+        }
```
brianhatchl: what's this doing?

brianhatchl (changes requested):
Can you provide a write up of how the multiselect folder behavior operates?
What operations are available for multiple selected folders
what ones are hidden.
What is the behavior of selecting a combo of folder and dataset?  Is this disallowed?
It would be great to write tests for the documented behavior

modules/Hoot/tools/folderTree.js
jackgrossman18: removed else

modules/Hoot/tools/folderTree.js
jackgrossman18: changed method name to better reflect purpose

modules/Hoot/tools/folderTree.js
jackgrossman18: thought if multiple folders are being selected, it may make sense to only allow the use to move them, if they want to rename a folder, they will have to select them individually

modules/Hoot/tools/folderTree.js
jackgrossman18: I did this to prevent duplication of code which listens for a ctrl key event on a folder

modules/Hoot/tools/folderTree.js
jackgrossman18: fixed

modules/Hoot/ui/managePanel/datasets.js
jackgrossman18: updated

modules/Hoot/ui/managePanel/datasets.js
jackgrossman18: updated

modules/Hoot/ui/modals/modifyFolder.js
jackgrossman18: Moving multiple folders ,and individual folders for that matter, we shouldn't have the option to move a folder and its children into the selected folder's parent.

We also shouldn't be capable of moving a selected folder or folders into a child of that folder. 

This method is attempting to present the user with a meaningful list of options for moving the selected folder.

jackgrossman18:
@brianhatchl I've made a few updates. 

Would it be a good idea add the option to both modify or move a folder for a single folder/dataset ?

brianhatchl:
Yes, it's not valid to attempt to move a folder into one of it's children.

brianhatchl (approved):

PR #1628, opened 2019-11-04
added option for importing single dataset on folder right click

jackgrossman18:
ref #1627


modules/Hoot/ui/managePanel/datasets.js
```diff
@@ -255,7 +255,13 @@ export default class Datasets extends Tab {
 
                 break;
             }
-            case 'importDatasets': {
+            case 'importDataset': {
+                let translations = await Hoot.api.getTranslations();
+                this.importSingleModal = new ImportDataset( translations, d.data ).render();
```
brianhatchl: Looks good, just sync this call with #1629 to be `d.data.path`

brianhatchl (approved):

PR #1629, opened 2019-11-04
updating path declaration in constructor of importDataset & importMulti

jackgrossman18:
ref #1626

### DIFF
--- a/modules/Hoot/ui/managePanel/datasets.js
+++ b/modules/Hoot/ui/managePanel/datasets.js
@@ -257,7 +257,7 @@ export default class Datasets extends Tab {
             }
             case 'importDatasets': {
                 let translations = await Hoot.api.getTranslations();
-                this.importMultiModal = new ImportMultiDataset( translations, d.data.name ).render();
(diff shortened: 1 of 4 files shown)


modules/Hoot/ui/managePanel/datasets.js
```diff
@@ -257,7 +257,7 @@ export default class Datasets extends Tab {
             }
             case 'importDatasets': {
                 let translations = await Hoot.api.getTranslations();
-                this.importMultiModal = new ImportMultiDataset( translations, d.data.name ).render();
+                this.importMultiModal = new ImportMultiDataset( translations, d.data ).render();
```
brianhatchl: why not just pass `d.data.path` here instead of `d.data.name`?
I don't see the advantage of passing the entire object.

brianhatchl (changes requested):

jackgrossman18:
@brianhatchl requested changes made

brianhatchl:
all changes required came across with #1628

PR #1642, opened 2019-11-06
Fixed #1641. grail pull stats not showing any ways when there should be

### LINKED ISSUE #1641: grail pull stats not showing any ways when there should be
```
[out:json][timeout:25];
// gather results
(
  // query part for: "building=*"
  node["building"]({{bbox}});
  way["building"]({{bbox}});
  relation["building"]({{bbox}});
);
// print results
out body;
>;
out meta;
```

bobi7:
ref #1641

### DIFF
--- a/modules/Hoot/tools/grailPull.js
+++ b/modules/Hoot/tools/grailPull.js
@@ -69,9 +69,7 @@ export default class GrailPull {
 
         this.loadingState();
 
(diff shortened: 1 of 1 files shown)


brianhatchl (approved):
LGTM :taco: 

wonder why we don't do this reformatting in services :thinking:

PR #1645, opened 2019-11-08
Fixed #1601. added checkboxes for canned overpass queries

### LINKED ISSUE #1601: add checkboxes for canned overpass queries
for work in #1589 we should also include a set of checkboxes that let users quickly make their overpass queries include things like buildings, roads, ect...

bobi7:
ref #1601

### DIFF
--- a/modules/Hoot/tools/overpassQueryPanel.js
+++ b/modules/Hoot/tools/overpassQueryPanel.js
@@ -78,11 +78,13 @@ export default class OverpassQueryPanel {
             .append( 'label' )
             .text( 'Custom Overpass query' )
                 .append( 'input' )
+                .attr( 'id', 'customQueryToggle' )
                 .attr( 'type', 'checkbox' )
                 .property( 'checked', checkboxStatus )
                 .on('click', () => {
                     const isChecked = checkboxLabel.property( 'checked' );
                     customQueryInput.classed( 'hidden', !isChecked );
+                    this.queryOptions.classed( 'hidden', !isChecked );
 
                     if ( isChecked ) {
                         this.submitButton.select( 'span' ).text( 'Next' );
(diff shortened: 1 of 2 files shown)


modules/Hoot/tools/overpassQueryPanel.js
```diff
@@ -122,11 +124,57 @@ export default class OverpassQueryPanel {
...
+        options.forEach( option => {
+            this.queryOptions.append( 'label' )
+                .text( option )
+                .append( 'input' )
+                .attr( 'id', option )
+                .attr( 'type', 'checkbox' )
+                .on( 'click', () => {
+                    this.queryBuilder();
+                } );
+        } );
```
brianhatchl: Could this be done the d3 way with `.data().enter().append()`?
Then the bound data could be used to drive the queryBuilder below by retrieving a tag term in a selection.
It should also clear out the query if no boxes are checked.

brianhatchl (changes requested):

modules/Hoot/tools/overpassQueryPanel.js
```diff
@@ -132,46 +132,51 @@ export default class OverpassQueryPanel {
...
-
+            } );
     }
 
     queryBuilder() {
-        let queryString = '[out:json][bbox:{{bbox}}];\n' +
+        const checkedInput = this.queryOptions.selectAll( 'input' )
+            .filter( function() {
+                return d3.select( this ).property( 'checked' ) === true;
+            });
```
brianhatchl: this can be done with the selector `input[type=checkbox]:checked`

brianhatchl (changes requested):

brianhatchl (approved):

PR #1646, opened 2019-11-08
use folderPaths getter which will filter by Public visibility

### LINKED ISSUE #1525: Folder lists should filter based on Public Data button state
image

if Public Data is not shown, public folders not owned by the users should be filtered out

brianhatchl:
closes #1525


brianhatchl:
Drop down folder paths should now reflect the same Public Data visibility setting as the Datasets pane.

bobi7 (approved):
🏆

Issue #1639, opened 2019-11-05
New layers made with clip tool does not refresh automatically in the datasets list


ryanmcweeny:
When using the Clip tool to create a new layer from an existing layer file it is necessary to click Refresh Datasets for it to appear in the list of datasets. Otherwise it will not appear. However, importing a single file does cause the list to refresh automatically.

PR #1649, opened 2019-11-14
updating selected destination path for clip, & refresh for all folders

jackgrossman18:
ref #1639

### DIFF
--- a/modules/Hoot/tools/clipDataset.js
+++ b/modules/Hoot/tools/clipDataset.js
@@ -133,7 +133,7 @@ export default class ClipDataset {
                             .concat( d.combobox.filter( function( l ) { return l.id !== folderId; } ).sort() );
                         that.createFolderListCombo( d3.select( this ), d );
                         d3.select( this ).property( 'value', Hoot.folders.findBy( 'id', folderId).name );
-                        d3.select( this ).property( '_value', Hoot.folders.findBy( 'id', folderId ).id );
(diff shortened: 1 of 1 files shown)


modules/Hoot/tools/clipDataset.js
```diff
@@ -193,7 +193,7 @@ export default class ClipDataset {
             let row         = d3.select( `#row-${ mapId }` ),
                 datasetName = row.select( '.datasetName' ),
                 outputName  = row.select( '.outputName' ),
-                folderId    = row.select( '.outputPath').property( '_value' );
+                folderId    = Hoot.folders.findBy( 'path', row.select( '.outputPath' ).property( 'value' ) ).id;
```
brianhatchl: The `_value` property should have the folder id.  Was this change correcting something else?

jackgrossman18: Before this change, you couldn't change the destination path correctly. 

If the folder path provided when the modal became available would be the folder path the clipped dataset was sent to, regardless of what you had selected the path to be.

brianhatchl: Looking elsewhere in the code, I think it should be `.attr('_value')`
that might fix it.

brianhatchl (changes requested):
There is one more to be replaced on line 136.

brianhatchl (approved):

PR #1651, opened 2019-11-14
Remove root as import path option for import multi & import dataset

jackgrossman18:
ref #1650

### DIFF
--- a/modules/Hoot/managers/folderManager.js
+++ b/modules/Hoot/managers/folderManager.js
@@ -79,7 +79,7 @@ export default class FolderManager {
     }
 
     filterVisible(f) {
-        return JSON.parse(Hoot.context.storage( 'publicVisibility' )) && f.id !== 0 || f.userId === Hoot.user().id && f.id !== 0;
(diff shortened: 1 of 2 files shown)


modules/Hoot/managers/folderManager.js
```diff
@@ -79,7 +79,7 @@ export default class FolderManager {
     }
 
     filterVisible(f) {
-        return JSON.parse(Hoot.context.storage( 'publicVisibility' )) || f.userId === Hoot.user().id;
+        return JSON.parse(Hoot.context.storage( 'publicVisibility' )) && f.id !== 0 || f.userId === Hoot.user().id && f.id !== 0;
```
maxgrossman: Just a suggestion, what if we put `f.id !== 0` at the beginning of this?
I am pretty sure that would achieve the same and I think doing so also says 'first and foremost, ignore the root folder'.

maxgrossman (approved):

modules/Hoot/managers/folderManager.js
brianhatchl: I don't think we can do it like this because we do need root for Move Folder (which uses filterVisible too)

Issue #1214, opened 2018-09-04
iD shows no error if Element Merge Service is not available


bwitham:
Saw this in the VGI instance where the Element Merge Service was not operational for some reason.  iD sends the merge request during the review session to the service but never reports back an error when no response comes back from it and the POIs never merge.

Issue #1006, opened 2018-03-06
Poi/poly merging incorrect arrow behavior


bwitham:
When showing POI/Poly reviews to the user:
* The review arrow should always point from POI to polygon.
* The user should not have the option to reverse the arrow.

bwitham:
my bad...its already in there...knew it had to be.  Just wasn't seeing the button.

bwitham:
Will repurpose this...

Issue #1635, opened 2019-11-05
Using undo after a merge does not bring back the merge button


ryanmcweeny:
When resolving issues in the conflation tool I have found that if you mistakenly merge two features and then undo the merge the merge button does not reappear as seen in the two images below.

![68161732-819d8380-ff24-11e9-93ea-f397f4507531](https://user-images.githubusercontent.com/24612000/68232868-068fa800-ffcc-11e9-8784-597736522c6b.png)
![68161733-819d8380-ff24-11e9-95d8-2cfb929488d0](https://user-images.githubusercontent.com/24612000/68232869-068fa800-ffcc-11e9-9d9f-139e66454967.png)

Issue #1480, opened 2019-07-03
Tags in review tables don't get updated after a review merge until after the "Resolved" button is clicked


bwitham:
Saw this with poi/poly conflation while working on https://github.com/ngageoint/hootenanny/issues/3322. If you merge two features during a review, the tags for the feature being merged into don't get updated in the red/blue tables until you click the "Resolved" button. It seems like the tag tables should be updated immediately after clicking the "Merge" button, but that's arguable. I think this has led to some user confusion that tags aren't merging correctly.

jackgrossman18:
cc@bwitham & @maxgrossman 

when the user clicks the merge button, which do you think would be the best change to make to the tag table. 

1.  Update the table so the new merge tag column overtakes the 'from' tags column as that feature is deleted. This will result in one large red table containing the new merged tags.

2. Only merge the 'from' tags into the 'to' tags column, leaving the 'from' tags column in the UI

3. Delete the entire tag table once the merge button is clicked.

bwitham:
That's a good question... 1) seems to make the most sense to me. 

It would be nice then after that if you clicked the undo button, the red and blue tables would reappear and the features be unmerged...but that may be difficult to do.

jackgrossman18:
Agreed, option 1 seems like the best choice. 
I'll work through your suggestion for the undo button action to see how much work it would take. Thanks

PR #1654, opened 2019-11-18
once conflict review feature is unmerged w/ undo reintroduce merge button

jackgrossman18:
ref #1635


maxgrossman:
This worked for sure when I clicked merge, then undo back to before all the actions within a merge (change tags, delete).

However, if I use undo/redo to go to the head and base of the undo/redo stack (click undo or redo until I cannot anymore in either direction), the button gets all out of sync.

Perhaps we can listen to the undone/redone events fired by the history class?
The class is accessible on context so it'd be avail in this class.

I think you'd be able to say something like...

on `undone`, if the thing that is being undone (via the annotation on the graph) is changed tags, we know we can show the merge button. This is because the first step of the merge is to change the tags of the target feature.

on `redone` If the thing being redone is that then you'd be able to say ok hide the merge button. (since that's the inverse of what we care about for undone)

maxgrossman (changes requested):
Nice! I added a comment. This seems to work in one direction (when I click undone).
We'll need this to be bi-directional (undoing and redoing), and I added an idea of how we might be able to achieve it.

brianhatchl:
@bobi7 I made the changes suggested by @maxgrossman.  Rather than checking for what change the unde/redo is peforming, it seems sufficient to just check the review features and make sure they are eligible for merge.  The test at play here is that both review features exist in the graph, which is not true after merge, but becomes true again when the merge is undone.

brianhatchl:
should also fix #1006 #1480

brianhatchl:
might also add a fix for #1476

modules/Hoot/ui/conflicts/info.js
```diff
@@ -26,19 +27,35 @@ export default class ConflictMetadata {
...
-        let colData    = this.data.currentFeatures,
-            tags1      = this.filterTags( colData[ 0 ] ? colData[ 0 ].tags : {} ),
+        let currentRelation;
+        try {
+            currentRelation = this.instance.graphSync.getCurrentRelation();
+        } catch (e) {
+            //current relation not found which is ok
+        }
+        //if triggered by history change and not in review mode return
+        if (!currentRelation) return;
```
brianhatchl, line 47: @jackgrossman18 because this method listens to history changes, it will be called outside the conflation review context and so getting the current relation may throw an "entity not found" error.

modules/Hoot/ui/conflicts/info.js
```diff
@@ -26,19 +27,35 @@ export default class ConflictMetadata {
...
+
+        let features = this.data.currentFeatures,
+            feature = Hoot.context.hasEntity( features[0].id ),
+            againstFeature = Hoot.context.hasEntity( features[1].id );
+
+        let colData = [feature, againstFeature];
```
brianhatchl, line 53: @jackgrossman18 this seems to me a simpler way to handle if a feature has been merged and 1) had tags updated or 2) been deleted.  By grabbing the features from the graph instead of the stale `currentFeatures` the tag table will now update correctly anytime review merges happen.

modules/Hoot/ui/conflicts/map.js
```diff
@@ -58,17 +58,7 @@ export default class Map {
...
-        if ( relation.tags[ 'hoot:review:type' ] === 'POI to Polygon' ||
-            ((feature && againstFeature) && feature.id.charAt( 0 ) === 'n' && againstFeature.id.charAt( 0 ) === 'n')
-        ) {
-            //TODO: check for poly and make that against feature if other is node
-            this.instance.merge.toggleMergeButton( false );
-            this.instance.merge.activateMergeArrow( feature, againstFeature );
-        } else {
-            this.instance.merge.toggleMergeButton( true );
-        }
+        this.instance.merge.checkMergeButton();
```
brianhatchl, line 61: @jackgrossman18 move this logic inside `merge.js`

modules/Hoot/ui/conflicts/merge.js
```diff
@@ -23,13 +25,25 @@ export default class Merge {
...
+
+        let that = this;
+        Hoot.context.history()
+            .on('undone.review_merge', function() {
+                that.checkMergeButton.call(that);
+            });
+        Hoot.context.history()
+            .on('redone.review_merge', function() {
+                that.checkMergeButton.call(that);
+            });
```
brianhatchl, line 46: @jackgrossman18 listen for undo/redo and call the same function that is called when a new review is loaded.

modules/Hoot/ui/conflicts/merge.js
```diff
@@ -38,28 +52,20 @@ export default class Merge {
      * @returns {Promise<void>}
      */
     async mergeFeatures() {
-        let features = _clone( this.data.currentFeatures ),
-            reverse  = d3.event.ctrlKey || d3.event.metaKey || this.mergeArrow.reverseFeatures,
-            featureUpdate,
-            featureDelete,
+        let reverse  = d3.event.ctrlKey || d3.event.metaKey && !this.isPoiPoly,
+            featureUpdate = _cloneDeep((reverse) ? this.mergeArrow.from : this.mergeArrow.to),
+            featureDelete = _cloneDeep((reverse) ? this.mergeArrow.to : this.mergeArrow.from),
```
brianhatchl, line 57: @jackgrossman18 we have to deep clone here because we modify the feature tags on line 68.  If we do a merge, then undo it and reverse it the element merge service errors because both featuers have the `hoot:merge:target=yes` tag.

brianhatchl: Also when checking if reverse merge is valid, we disallow if it is POI->Poly.  We can simply disallow because code in `checkMergeButton` ;omr 358 makes sure the poi and poly features are always in the right order.

modules/Hoot/ui/conflicts/merge.js
```diff
@@ -328,11 +314,23 @@ export default class Merge {
             return;
         }
 
-        let pt1   = d3.geoCentroid( this.mergeArrow.to.asGeoJSON( Hoot.context.graph() ) ),
-            pt2   = d3.geoCentroid( this.mergeArrow.from.asGeoJSON( Hoot.context.graph() ) ),
+        let pt1   = d3.geoCentroid( this.mergeArrow.from.asGeoJSON( Hoot.context.graph() ) ),
+            pt2   = d3.geoCentroid( this.mergeArrow.to.asGeoJSON( Hoot.context.graph() ) ),
```
brianhatchl, line 325: I found that the original code was putting the arrow direction in the opposite from the intended from/to direction

modules/Hoot/ui/conflicts/merge.js
```diff
@@ -342,4 +340,34 @@ export default class Merge {
 
         Hoot.context.background().updateArrowLayer( gj );
     }
+
+    checkMergeButton() {
+        let features = _clone( this.data.currentFeatures ),
+            toFeature = Hoot.context.hasEntity( features[0].id ),
+            fromFeature = Hoot.context.hasEntity( features[1].id ),
```
brianhatchl, line 354: Use dynamic features retrieved from graph.  This allows the show/hide merge button to work on undo/redo.

modules/Hoot/ui/conflicts/merge.js
```diff
@@ -342,4 +340,34 @@ export default class Merge {
...
+            relation = this.instance.graphSync.getCurrentRelation();
+
+        if ( toFeature && fromFeature) {
+            this.isPoiPoly = (toFeature.type === 'node' && fromFeature.type === 'way')
+                || (toFeature.type === 'way' && fromFeature.type === 'node');
```
brianhatchl, line 359: We need to define POI->Poly merge this way (instead of using `hoot:review:type` because after a merge, other review relations will be updated to refer to the merged element.  After a poi->poly merge the referred element may no long be a POI and may be two polys but the review type remains unchanged.  This may change when we implement this long ago ticket #1003 to allow poly->poly merges.

modules/Hoot/ui/conflicts/merge.js
```diff
@@ -342,4 +340,34 @@ export default class Merge {
...
+
+            let isPoiPoi = (toFeature.type === 'node' && fromFeature.type === 'node');
+
+            if (this.isPoiPoly || isPoiPoi) {
+                this.toggleMergeButton( false );
+                if (this.isPoiPoly) {
+                    let poi = (toFeature.type === 'node') ? toFeature : fromFeature;
+                    let poly = (toFeature.type === 'way') ? toFeature : fromFeature;
+                    this.activateMergeArrow( poi, poly ); //always merge poi->poly
```
brianhatchl, line 368: Make sure the feature order for the merge array is always poi->poly

modules/Hoot/ui/conflicts/merge.js
```diff
@@ -38,28 +52,20 @@ export default class Merge {
      * @returns {Promise<void>}
      */
     async mergeFeatures() {
-        let features = _clone( this.data.currentFeatures ),
-            reverse  = d3.event.ctrlKey || d3.event.metaKey || this.mergeArrow.reverseFeatures,
-            featureUpdate,
-            featureDelete,
+        let reverse  = d3.event.ctrlKey || d3.event.metaKey && !this.isPoiPoly,
```
bobi7: should we wrap the `d3.event.ctrlKey || d3.event.metaKey` in parenthesis for readability?

brianhatchl:
Also fixes #1214

brianhatchl:
and fixes #1795

bobi7 (approved):

PR #1658, opened 2019-11-25
make sure layer ids are numbers when we load layers

### LINKED ISSUE #1657: findByLoaded id output is string when findBy input wants number
I just tried to test out running a conflation and the below javascript error was thrown

```
Uncaught TypeError: Cannot read property 'bbox' of undefined
    at LayerConflate.preConflation (layerConflate.js:218)
    at LayerConflate.handleSubmit (layerConflate.js:255)
    at HTMLButtonElement.<anonymous> (layerConflate.js:90)
    at HTMLButtonElement.<anonymous> (on.js:27)
```

In [this block of code](https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/Hoot/ui/sidebar/layerConflate.js#L179-L210) I noticed that the `data.INPUT_1` used on line 218 is a number as a string. I tried running the same findBy function with the id as an actual number and it returned the config.

As to why the id is a string, I think it is because of 179 (where we get the layer config). It uses findLoadedBy, which looks at the layer manager's loadedLayers map. 
This map's configs appear to have the layer ids as strings.

Perhaps we just make that a number and not a string.

maxgrossman:
closes #1657

### DIFF
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -211,7 +211,7 @@ export default class Layers {
 
             let layer = {
                 name: params.name,
-                id: params.id,
+                id: Number(params.id),
                 refType: params.refType,
                 color: params.color,
                 isMerged: params.isMerged || false,
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -196,12 +196,16 @@ export default class Layers {
     }
 
     async loadLayer( params, skipCheckForReviewsAndZoom ) {
+        //Some type checking
+        if (typeof params.id !== 'number') {
+            throw 'params.id must be a number';
+        }
         try {
             let mapId       = params.id,
                 tags        = await this.hoot.api.getMapTags( mapId ),
                 layerExtent;
 
-            let lyr = this.findBy( 'id', +mapId);
+            let lyr = this.findBy( 'id', mapId);
             if (lyr.bbox) {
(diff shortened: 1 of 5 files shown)


bobi7 (approved):
🥇

brianhatchl:
:eyes: This gives me pause as there's a ton of places we filter loadedLayers by id and I'm not sure all of them are using a number.

brianhatchl:
I think we need to update this line to coerce a number
https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/osm/entity.js#L53

brianhatchl:
Actually, I think string coercion happens automatically when using square bracket notation to reference a key in a js object, so the above may not matter.
e.g. `foo[1]` and `foo["1"]` reference the same thing
I'm still curious which line of code is sending a string id in params to loadLayer()

maxgrossman:
#dontknow, I can debug more and get a better answer about where we pass a string @brianhatchl

brianhatchl:
I'm on it and have found the origins of the string id.  Going to add some more defensive coding, there are many places we call `layers.findBy( 'id'` and don't coerce input.

maxgrossman:
I see that `selectedLayer` can be used sometimes in layerAdd class.
This thing uses d3 attr func which returns a string.

brianhatchl:
I threw some commits in to throw an exception if we call loadLayer with an id param that's not number
and made sure to cast it where necessary.  I'd rather we do this than paper over sloppy code.

PR #1659, opened 2019-11-26
root folder should be available to modify folder, no others need it

brianhatchl:


modules/Hoot/tools/clipDataset.js
```diff
@@ -231,7 +231,7 @@ export default class ClipDataset {
                 } )
                 .then( resp => {
```
jackgrossman18: i think the `resp` needs to be wrapped in `async ()`

jackgrossman18 (changes requested):

PR #1667, opened 2019-12-04
copy the array before reverse

brianhatchl:
The column counts had the wrong values because the array was reversed in place.


bobi7 (changes requested):
🚀

bobi7 (approved):

PR #1670, opened 2019-12-05
Fixed #1662. add admin panel sorting and filtering

### LINKED ISSUE #1662: add admin panel sorting and filtering
add sorting to admin panel columns for display name and last authorized
add filter to the privileges column

bobi7:
ref #1662


brianhatchl (approved):
added right-click contextmenu handler to filterable, but :100:

Issue #1655, opened 2019-11-18
pass campaign changeset comment to hoot


brianhatchl:
iD gets a changeset comment in the url hash.  We should pass this to the apply (upload) changeset command.

brianhatchl:
use `comment`, ~~`hashtags`~~, ~~`source`~~ as changeset tags if provided by the campaign.

PR #1672, opened 2019-12-05
add UI controls for grail changeset comment, hashtags, source

brianhatchl:
refs #1655


modules/Hoot/managers/api.js
```diff
@@ -1227,22 +1225,23 @@ export default class API {
             } );
     }
 
-    differentialStats( jobId, includeTags ) {
+    changesetStats( jobId, includeTags ) {
```
brianhatchl, line 1228: renamed a bunch of method from "differential" to "changeset" to reflect that it's not only for the differential changeset workflow

modules/ui/commit.js
```diff
@@ -257,21 +257,21 @@ export function uiCommit(context) {
...
+        //     .attr('class', 'modal-section tag-section raw-tag-editor')
+        //     .merge(tagSection);
+
+        // var expanded = !tagSection.selectAll('a.hide-toggle.expanded').empty();
+        // tagSection
+        //     .call(rawTagEditor
+        //         .expanded(expanded)
+        //         .readOnlyTags(readOnlyTags)
+        //         .tags(_clone(_changeset.tags))
+        //     );
```
brianhatchl, line 274: commented this upstream section out because we don't ever show tags on hoot changesets

modules/Hoot/ui/modals/changesetStats.js
```diff
@@ -0,0 +1,226 @@
...
+                    Hoot.context.storage('commentDate', Date.now());
+                }
+            }
+
+            that.tags = Object.assign(that.tags, changed);
+
+            that.changesetSection
+                .call(that.changesetEditor
+                    .tags(that.tags)
+                );
```
brianhatchl: this is the meat of the PR, adding the changeset editor used in commit.js to the apply changeset workflow:

image

modules/Hoot/managers/api.js
```diff
@@ -738,7 +738,14 @@ export default class API {
             .then( resp => {
                 let fileBlob = new Blob( [ resp.data ], { type: 'application/xml' } );
                 saveAs( fileBlob, `changeset_${id}.osc` );
-            });
+            })
+            .catch( err => {
+                console.log(err);
```
bobi7: remove console.log

modules/Hoot/ui/modals/changesetStats.js
```diff
@@ -0,0 +1,226 @@
...
+import { uiChangesetEditor } from '../../../ui/changeset_editor';
+
+export default class ChangesetStats {
+    constructor( job, data ) {
+        this.job = job;
+        this.changesetInfo = data;
+        this.includeTags = false;
+        this.changesetEditor = uiChangesetEditor(Hoot.context)
+            .on('change', changeTags);
+        var that = this;
```
bobi7: change to const just for consistency of us not using var

modules/Hoot/managers/api.js
brianhatchl: I added this because the promise catch was silently swallowing js errors in addition to xhr errors

bobi7: how about console.error since i see us using that in cases like this?

bobi7 (approved):
🥇 💯

PR #1684, opened 2019-12-13
fixed #1683. user without privileges was causing ui to not be usable

### LINKED ISSUE #1683: brand new users without privileges cannot do anything in the UI
I was testing out the UI as a brand new dummy user and realized I was not able to do anything after logging in. If I cleared my cache and logged in again, I can use the UI as I expect. I created a second dummy user; the same behavior.

After investigating, I think this is what is going on.

The first time you try to login as a fresh user not in hoot users table, no record is written about you the fresh user in the privliges table (or row, sorry for being iffy on schema).

I claim this based on this [upsert](https://github.com/ngageoint/hootenanny/blob/master/hoot-services/src/main/java/hoot/services/controllers/auth/UserManagerImpl.java#L173-L180) function. This gets called when oAuth makes a request to the `/verify` route (the one in between logging in with the OSM site and getting sent back to hoot).

This leads the user object seen by the ui to look like this.

```
{
   "email": "...",
   "id": "...",
   "display_name":"...",
   "last_authorized": "...",
   "provider_created_at": "..."
}
```

It is missing the "privileges" entry. My understanding, per that upsert function's logic, is that only when you log in after being in the user table will the privileges be set and that user object have the priv entry we need.

The effect of this, the bug, I think points to this line [here](https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/Hoot/hoot.js#L95-L97). We try to access that entry for fresh users, it doesn't exist, javascript barfs and the user cannot do anything.

here's a snapshot of my console during the bug.
Screenshot from 2019-12-13 09-04-46

Minifying prevents knowing the line number, but a grep on `advanced` in the source code returns the lines above.

bobi7:
ref #1683

### DIFF
--- a/modules/Hoot/hoot.js
+++ b/modules/Hoot/hoot.js
@@ -94,7 +94,7 @@ class Hoot {
     async getGrailMetadata() {
         const privileges = this.user().privileges;
 
-        if ( privileges.advanced && privileges.advanced === 'true' ) {
+        if ( privileges && privileges.advanced && privileges.advanced === 'true' ) {
             const { data } = await this.api.grailMetadataQuery();
             this.config.referenceLabel = data.railsLabel;
             this.config.secondaryLabel = data.overpassLabel;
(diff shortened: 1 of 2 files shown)


maxgrossman (approved):
safe checking all star

PR #1688, opened 2019-12-16
fixed #1663. add last accessed info to folders in the datasets panel

### LINKED ISSUE #1663: add last accessed, created, size info to folders in the datasets panel

bobi7:
ref #1663

### DIFF
--- a/modules/Hoot/ui/modals/exportData.js
+++ b/modules/Hoot/ui/modals/exportData.js
@@ -85,7 +85,7 @@ export default class ExportData {
             const children = folder.children || folder._children || (folder.data ? folder.data._children : []) || [];
 
             // skip if no children
-            if (!children.length === 0) continue;
+            if (children.length === 0) continue;
 
             totalSize += children.filter( child => child.size ).reduce( ( acc, dataset ) => {
                 return acc + dataset.size;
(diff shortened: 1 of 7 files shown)


modules/Hoot/ui/modals/exportData.js
```diff
@@ -85,7 +85,7 @@ export default class ExportData {
             const children = folder.children || folder._children || (folder.data ? folder.data._children : []) || [];
 
             // skip if no children
-            if (!children.length === 0) continue;
+            if (children.length === 0) continue;
```
brianhatchl, line 88: Wow, is this a bug fix?

brianhatchl (changes requested):
Noticed a subtle formatting difference between folders and datasets:
Screenshot from 2019-12-17 14-29-45
We might even change this very precise DTG format for something easier to read (who needs seconds and milliseconds).  Maybe re-label `Import Date` to `Create Date`?
Also, some of my folders aren't getting the rollup size calculated accurately:
Screenshot from 2019-12-17 14-33-35

modules/Hoot/ui/modals/exportData.js
bobi7: yea found it in another place too. webstorm catching mistakes!

brianhatchl (approved):
:kick_scooter: let's go

Issue #1682, opened 2019-12-12
Expose single input derive changeset command to the ui


maxgrossman:
looking at the [derive changeset docs](https://github.com/ngageoint/hootenanny/blob/5b6a92e084a324a7b7946ee03119f7de7ca3bc2b/docs/commands/changeset-derive.asciidoc), the command can be given a single input and (I assume) make an osc with a bunch of adds. If data is sparse in an area or I know my secondary source does not overlap existing data, it'd perhaps be nice if I could take an existing map and run a changeset-derive command that generates an adds only osc file.

Do this with caution! might even be something to make for users with only certain privileges.

PR #1690, opened 2019-12-18
add derive changes [adds only] action

brianhatchl:
refs #1682

### DIFF
--- a/modules/Hoot/tools/grailPull.js
+++ b/modules/Hoot/tools/grailPull.js
@@ -254,10 +254,11 @@ export default class GrailPull {
                     let submitPromises = [];
 
                     // Finding layer id by name is fine here because we check for duplicate name in the grail pull
-                    if ( referenceCheckbox ) {
+                    let refLayer = Hoot.layers.findBy( 'name', railsParams.input1 );
+                    if ( referenceCheckbox && refLayer ) {
                         let refParams = {
                             name: railsParams.input1,
-                            id: Hoot.layers.findBy( 'name', railsParams.input1 ).id,
+                            id: refLayer.id,
                             color: 'violet',
                             refType: 'primary'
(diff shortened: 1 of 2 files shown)


modules/Hoot/tools/grailPull.js
```diff
@@ -254,21 +254,23 @@ export default class GrailPull {
...
                         submitPromises.push( Hoot.ui.sidebar.forms.reference.submitLayer( refParams ) );
                     }
 
-                    if ( secondaryCheckbox ) {
+                    let secLayer = Hoot.layers.findBy( 'name', overpassParams.input1 );
+                    if ( secondaryCheckbox && secLayer ) {
                         let secParams = {
                             name: overpassParams.input1,
-                            id: Hoot.layers.findBy( 'name', overpassParams.input1 ).id,
+                            id: secLayer.id,
```
brianhatchl, line 273: fixes npe when grail pull fails for a layer

modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -691,7 +691,19 @@ export default class Jobs extends Tab {
...
+                                icon: 'add_to_photos',
+                                action: async () => {
+                                    const params  = {};
+                                    params.input1 = d.mapId;
+                                    params.parentId = d.jobId;
+
+                                    Hoot.api.deriveChangeset( params )
+                                        .then( resp => Hoot.message.alert( resp ) );
+                                }
+                            });
```
brianhatchl, line 705: Adds a new action to do single input derive changeset [adds only]
image
_*third from left_

It should be available to any import or conflate output that derive changeset replacement is, but doesn't need a reference layer to go with it.

bobi7 (approved):
🔥

PR #1697, opened 2020-01-07
Fixed #1691. deleting folder doesnt cause datasets inside to get deleted

### LINKED ISSUE #1691: deleting folder doesnt cause datasets inside to get deleted as well

bobi7:
ref #1691


bobi7:
the issue had to do with when a open folder gets deleted, the data object is sent.
if you inspect the folder dom in the datasets panel and look at the properties you will see __data__ and inside there is a data object. The issue is that if the folder was open the children variable inside the `data` varialbe would be set to null but the one under __data__ would be set to the proper children.

brianhatchl (approved):
:football: you like that!

PR #1700, opened 2020-01-09
Add jobs action to download conflicted osc file after upload

brianhatchl:
refs #1687

### DIFF
--- a/modules/Hoot/managers/api.js
+++ b/modules/Hoot/managers/api.js
@@ -727,9 +727,9 @@ export default class API {
             });
     }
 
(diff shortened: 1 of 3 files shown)


brianhatchl:
This PR adds an action to download the conflicted changes in an osc file for upload changeset jobs image

It also is part of https://github.com/ngageoint/hootenanny/pull/3696 that makes changes to the grail pull end point to allow a dataset to be refreshed so this can be invoked from the UI instead of being an additional command in the upload job.

bobi7 (approved):
LGTM ⏫

Issue #1702, opened 2020-01-13
Issues loading in custom basemaps


ryanmcweeny:
After uploading a tif to be used as a custom basemap the image will not load on the map. Furthermore when attempting to delete and upload the layer again Hoot is keeping all previous iterations of the layer and not deleting them.
<img width="1278" alt="Screen Shot 2020-01-13 at 3 04 02 PM" src="https://user-images.githubusercontent.com/24612000/72288306-c3c6fd80-3616-11ea-9f66-f32433ec0e14.png">

PR #1706, opened 2020-01-15
basemaps that are processing, will be disabled w/in basemap modal

jackgrossman18:
ref #1702 & #1694


jackgrossman18:
this pr can replace:
https://github.com/ngageoint/hootenanny-ui/pull/1695

maxgrossman:
status_messages

I noticed that the basemaps upload for 'processing' basemaps shows up even when I am not on the basemaps tab. I also wonder if we can just hide those that are 'processing'. The jobs panel should show the job process going on until it is uploaded + it includes things like percent complete.

modules/Hoot/ui/managePanel/basemaps.js
```diff
@@ -52,18 +53,38 @@ export default class Basemaps extends Tab {
...
+                basemap.unavailable.push(d);
+            }
+        });
+        return basemap;
+    }
+
     populateBasemaps( basemaps ) {
         let instance = this;
 
+        let allBasemaps = this.checkBasemapStatus( basemaps );
```
maxgrossman: Related to the GIF i posted, maybe here it just is a filter so that only those that match `d.status==='enabled' || d.status==='disabled'` are returned

jackgrossman18:
@maxgrossman check created for empty job commands

jackgrossman18:
@maxgrossman

maxgrossman (changes requested):
I think for the purpose of only showing basemaps that are already processed (so those with statuses enabled or disabled), the only you have added the basemaps.js is all that we need.

That said, just the one comment and removing the code from the other files.
There is still the piece to figure out how to get percentage complete from the gdal->tiles command. That should be done in a separate ticket I think.

modules/Hoot/ui/managePanel/basemaps.js
```diff
@@ -52,18 +53,33 @@ export default class Basemaps extends Tab {
...
 
             this.populateBasemaps( basemaps );
+
         } catch ( e ) {
             window.console.error( 'Unable to retrieve basemaps' );
             throw new Error( e );
         }
     }
 
+    checkBasemapStatus(basemaps) {
```
maxgrossman: Rather than a forEach to make a new list, what about a filter that returns when the condition you have in the if statement on line 67 is true.

maxgrossman: and maybe not even a new function, just do that filter on line 78

maxgrossman (approved):

PR #1713, opened 2020-01-23
fixed #1668. fixed bug with pressing enter with replace dropdown open

### LINKED ISSUE #1668: unexpected changeset replace behavior
if you open the dropdown for changeset replace and you press enter instead of selecting an option the job gets run and it doesnt appear to auto select the item in the dropdown either

steps:
grail pull
click changeset replace button in jobs panel
click to open the dropdown but dont select anything. instead press enter

bobi7:
ref #1668

### DIFF
--- a/modules/Hoot/ui/modals/grailDatasetPicker.js
+++ b/modules/Hoot/ui/modals/grailDatasetPicker.js
@@ -27,7 +27,6 @@ export default class GrailDatasetPicker {
                     readonly: 'readonly',
                     sort: false,
                     itemKey: 'name',
-                    _value: this.layer.id,
                     _valueKey: 'id',
                     onChange: d => this.handleSubmit( d )
                 }]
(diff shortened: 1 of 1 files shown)


modules/Hoot/ui/modals/grailDatasetPicker.js
```diff
@@ -27,7 +27,6 @@ export default class GrailDatasetPicker {
                     readonly: 'readonly',
                     sort: false,
                     itemKey: 'name',
-                    _value: this.layer.id,
```
brianhatchl, line 30: How come you removed this?  `_value` is the hack to provide the combobox with a value that differs from the option label.

bobi7: it was giving it a default value which would be the layer itself so when you open the combobox and press enter, even though you didnt select an option it would go to the submit function call, retrieve the _value, and get that default.

The combobox still looks the same as far as I can see

brianhatchl (approved):

PR #1716, opened 2020-01-27
1712

jackgrossman18:
ref #1712

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2209,6 +2209,7 @@ div.combobox {
     white-space: normal !important;
     overflow-wrap: break-word;
     word-wrap: break-word;
+    padding: 0px 5px 0px 5px;
 }
 
 .adv-top {
@@ -2220,17 +2221,12 @@ div.combobox {
 }
 
 .hoot-field-label-button {
-    height: 100%;
-    border: 1px solid #ccc;
-    border-radius: 0 3px 3px 0;
-    position: absolute;
-    top: 0;
-    right: 0;
-    width: 32px;
-    height: 32px;
-    background: #f6f6f6;
-    border-left-width: 1px;
-    border-right-width: 0;
+        border: 1px solid #ccc;
+        border-radius: 0 3px 3px 0;
+        height: inherit;
+        width: 10%;
+        background: #f6f6f6;
+        border-left-width: 1px;
 }
 
 .checkbox-opt {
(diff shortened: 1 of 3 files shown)


modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -291,6 +294,49 @@ export default class AdvancedOpts {
...
+                            this.showing = false;
+                        }
+                        else {
+                            if (d.input === 'checkbox') {
+                                getDescId.classed('hoot-form-field-checkbox', false);
+                                getDescId.classed('hoot-form-field-checkbox-clicked', true );
+
+                            }
+                                getDescId
+                                    .append('div')
```
maxgrossman: image

There seems to be a state in which having 2 dropdowns open leads to adding divs over and over if you click on one of the info buttons.

I think we can avoid this if we don't append on click, but show a div that we add in the same way we do this button  (d3 data binding cycle).

More succinctly I think you...

1. add the description div, not in the callback, but right underneath where 339 following d3 data binding flow
2. add a class for hidden to the div (there should be one we have already in the ui), if not, a display = none
3. then in the callback, all you do is set the value of the class/display property that shows/hides the div as the inverse of what it is `selection.classed('name-of-class, !selection.classed('name-of-class')`

css/80_app.css
```diff
@@ -2198,9 +2198,55 @@ div.combobox {
...
+.adv-top {
+    margin-top: 10px;
+}
+
+.adv-opt-icon {
+    opacity: .5;
+}
+
+.hoot-field-label-button {
+        border: 1px solid #ccc;
```
maxgrossman: I think it looks fine without the border. adding the border creates the appearance of a double :rainbow:, oops i mean border!

maxgrossman (changes requested):
See the comments above.

jackgrossman18:
@maxgrossman requested changes made

maxgrossman:
This is some serious nitpick, sorry, but I also noticed that "active" css is being applied at times when we might not expect it to. 

I think maybe we could only apply a tint to the background color "on hover"
hover_css
.

jackgrossman18:
@maxgrossman

jackgrossman18:
@maxgrossman ready to merge :triumph:

maxgrossman (approved):
:cake:

PR #1725, opened 2020-01-31
cleaning opts that prevent adv opt panel functionality

jackgrossman18:
ref #1724


maxgrossman:
We can close this. 
[#3750](https://github.com/ngageoint/hootenanny/pull/3750) in hootenanny fixes this.

Issue #1723, opened 2020-01-31
UI job status log not showing all logging statements


bwitham:
The status log leaves out several status messages from a conflate job. This is making it more difficult to diagnose user problems when a job crashes, since we don't know exactly where its crashing.

@SeeBee2018

brianhatchl:
We can do something about this.  There is more info on command text and stdout available in the job status responses that we can optionally display.

bwitham:
Cool, thanks.

PR #1727, opened 2020-02-04
option to show verbose log with command string

brianhatchl:
refs #1723


brianhatchl:
Gonna merge so we get into nightly server for testing tomorrow.  Review still appreciated.

Issue #1726, opened 2020-02-03
Reset button in advanced options ui does nothing!


maxgrossman:
I probably am to blame for this, but the reset button in the advanced options ui does not reset any of the advanced options ui. I see this javascript error too!!

```
advancedOpts.js:426 Uncaught TypeError: group.exit is not a function
    at AdvancedOpts.createGroups (advancedOpts.js:426)
    at HTMLButtonElement.<anonymous> (advancedOpts.js:87)
    at HTMLButtonElement.<anonymous> (on.js:27)
createGroups @ advancedOpts.js:426
(anonymous) @ advancedOpts.js:87
(anonymous) @ on.js:27
```

maxgrossman:
another bug

when you click a label of an option title you get

```
Uncaught TypeError: Cannot read property 'classList' of null
    at classList (classed.js:6)
    at Selection.classed (classed.js:66)
    at HTMLSpanElement.<anonymous> (advancedOpts.js:206)
    at HTMLSpanElement.<anonymous> (on.js:27)
```

I believe this is because I used an arrow function in the click callback the err line is in then tried using the `this` object. That does not work with arrow functions, it probably needs to just be changed to a regular `function() {}` then error should go away.

PR #1730, opened 2020-02-05
adding advOpts to advOpt panel reset

jackgrossman18:
ref #1726

### DIFF
--- a/modules/Hoot/ui/sidebar/advancedOpts.js
+++ b/modules/Hoot/ui/sidebar/advancedOpts.js
@@ -8,6 +8,7 @@ import _cloneDeep from 'lodash-es/cloneDeep';
 import _map       from 'lodash-es/map';
 import _isEmpty   from 'lodash-es/isEmpty';
 import _isBoolean from 'lodash-es/isBoolean';
+import _forEach   from 'lodash-es/forEach';
 
 import { d3combobox } from '../../../lib/hoot/d3.combobox';
 import { svgIcon }    from '../../../svg';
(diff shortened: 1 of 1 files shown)


jackgrossman18:
@maxgrossman :triumph:

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -83,7 +84,14 @@ export default class AdvancedOpts {
             .append( 'button' )
             .classed( 'advanced-opts-reset button secondary strong', true )
             .text( 'Reset' )
-            .on( 'click', () => this.createGroups() );
+            .on( 'click', () => {
+                let hiddenOpts = [];
+                let getOpts = d3.selectAll('.group-body.fill-white.hidden').data();
```
maxgrossman: I think here you can just use that d3 `.each` function, and in that callback just push the .name property. Rather than making a new list out of the selection and looping over it

modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -414,7 +422,7 @@ export default class AdvancedOpts {
         notNumber.dispatch( isNumber ? 'mouseleave' : 'mouseenter' );
     }
 
-    createGroups(advOpts) {
+    createGroups(advOpts, hiddenOpts ) {
```
maxgrossman: Also here, as it pertains to the null check on 493.
Having a null check is ok, but I think if we can avoid null checks, that's desirable. 
To do that you could just make hiddenOpts parameter default to an empty list in the method definition.

So you would change `createGroups(advOpts, hiddenOpts)`  to `createGroups(advOpts, hiddenOpts = [])`

maxgrossman (changes requested):
Just the 2 comments.
Otherwise, code looks good.

maxgrossman (approved):
yay!

PR #1737, opened 2020-02-13
Fixed #1704. job panel offset bug

### LINKED ISSUE #1704: job panel offset bug
lets say you have a lot of jobs and are on page 3
Then someone deletes a lot of the jobs (you can do this in database so you dont have to switch users)
You will remain on page 3 when there isnt enough data

bobi7:
If the user is on a page with no data, it will go to the last page with data now

ref #1704


brianhatchl (approved):
I found one more condition I added a fix for
:vulcan_salute:

PR #1744, opened 2020-03-03
1743 multiuser conflation campaign folders

maxgrossman:
closes #1744 

added methods (that we use in grailPull now) to only look at folders that I own, and then when we add subfolders, make sure that when we match on parents, I also filter for only those folders I own.

Then also separated the logic for gpx/conflation campaign grailPull folders from regular grail pull folders.

Last, after @brianhatchl suggested, put the project/task names in session storage so we don't have to try to these (like we were before). Now, these will exist so long as you have the tab open to hoot


bobi7 (approved):
💯

Issue #1742, opened 2020-03-02
Add diff-remaining.osc to zip with diff-error.osc


brianhatchl:
if either of those files are present after uploading a changeset, make them available for download

PR #1745, opened 2020-03-03
save all osc files for upload changeset including

brianhatchl:
original osc, diff-error, diff-remaining, and debug osc traffic

refs #1742

### DIFF
--- a/modules/Hoot/ui/managePanel/jobs.js
+++ b/modules/Hoot/ui/managePanel/jobs.js
@@ -642,7 +642,30 @@ export default class Jobs extends Tab {
                             title: 'download conflicted changes',
                             icon: 'archive',
                             action: async () => {
-                                Hoot.api.saveChangeset( d.tags.parentId, 'diff-error' )
+                                var param = {
+                                    input: d.tags.parentId,
+                                    inputtype: 'changesets',
+                                    outputname: d.tags.parentId,
+                                    outputtype: 'zip'
+                                };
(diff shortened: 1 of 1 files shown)


modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -642,7 +642,30 @@ export default class Jobs extends Tab {
...
+                                    inputtype: 'changesets',
+                                    outputname: d.tags.parentId,
+                                    outputtype: 'zip'
+                                };
+                                // Hoot.api.saveChangeset( d.tags.parentId, 'diff-error' )
+                                //     .catch( err => {
+                                //         console.error(err);
+                                //         Hoot.message.alert( err );
+                                //         return false;
+                                //     } );
```
bobi7, line 656: remove?

modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -642,7 +642,30 @@ export default class Jobs extends Tab {
                             title: 'download conflicted changes',
                             icon: 'archive',
                             action: async () => {
-                                Hoot.api.saveChangeset( d.tags.parentId, 'diff-error' )
+                                var param = {
```
bobi7: we should use 'const' or 'let'

modules/Hoot/ui/managePanel/jobs.js
brianhatchl: we'll revert to this once debug troubleshooting is done I think, so leave it for now

bobi7 (approved):

PR #1747, opened 2020-03-09
get job error message

brianhatchl:
refs https://github.com/ngageoint/hootenanny/pull/3886


brianhatchl:
This uses the job error resource to reuse the logic for getting a UI appropriate error message for a failed job (be it internal or external command).

PR #1749, opened 2020-03-11
Fixed #1600. restored nodescount query to allow fetching above zoom 16

### LINKED ISSUE #1600: Restore the nodescount query to allow fetching data above zoom level 16
Hoot1 would query each visible tile for nodes count and load them if the total nodes count was less than a threshold which I think was 50k.  This allows viewing sparse datasets at zooms above 16.

bobi7:
ref #1600


brianhatchl:
superseded by #1975

Issue #1746, opened 2020-03-05
Jobs panel bug, refresh layers called endlessly


brianhatchl:
When the jobs history response is empty, the refresh layers call repeats endlessly

PR #1753, opened 2020-03-18
refreshLayers api call not made if jobsHistory api call returns empty

jackgrossman18:
ref #1746

### DIFF
--- a/modules/Hoot/ui/managePanel/jobs.js
+++ b/modules/Hoot/ui/managePanel/jobs.js
@@ -174,9 +174,9 @@ export default class Jobs extends Tab {
     }
 
     jobsHistoryTotalCheck(jobsHistory) {
-        let activeTotal = Hoot.ui.managePanel.active.total;
+        let activeTotal = this.total;
 
-        if ( activeTotal < jobsHistory.total ) {
+        if ( activeTotal && activeTotal < jobsHistory.total ) {
             return true;
         }
     }
--- a/modules/Hoot/ui/managePanel/jobs.js
+++ b/modules/Hoot/ui/managePanel/jobs.js
@@ -177,7 +177,7 @@ export default class Jobs extends Tab {
         let jobsRunning = await Hoot.api.getJobsRunning();
         let jobsHistory = await Hoot.api.getJobsHistory(this.params);
 
-        if ( jobsHistory.total > 0 && this.total ) {
+        if ( jobsHistory.total > this.total ) {
             await Hoot.layers.refreshLayers();
         }
         this.total = jobsHistory.total;
--- a/modules/Hoot/ui/managePanel/jobs.js
+++ b/modules/Hoot/ui/managePanel/jobs.js
@@ -173,10 +173,21 @@ export default class Jobs extends Tab {
         }
     }
 
+    jobsHistoryTotalCheck(jobsHistory) {
+        let activeTotal = Hoot.ui.managePanel.active.total;
+
+        if ( activeTotal < jobsHistory.total ) {
+            return true;
+        }
+    }
+
     async loadJobs() {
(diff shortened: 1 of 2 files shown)


brianhatchl (changes requested):
This doesn't fix the runaway, rapid fire request for me.  In my inspection I wasn't able to identify why the requests go wild, but they are for running and history jobs as well as layer requests.

Making this change prevented the runaway requests, but also caused the jobs history table not to render all the way:
```
@@ -187,14 +187,14 @@ export default class Jobs extends Tab {
 
         let checkTotal = this.jobsHistoryTotalCheck(jobsHistory);
 
         if ( jobsHistory.total > 0 && checkTotal ) {
             await Hoot.layers.refreshLayers();
+            this.total = jobsHistory.total;
+            this.paging.updatePages();
+            this.populateJobsHistory( jobsHistory.jobs );
         }
-        this.total = jobsHistory.total;
-        this.paging.updatePages();
-        this.populateJobsHistory( jobsHistory.jobs );
         this.populateJobsRunning( jobsRunning );
     }
 
     [getJobTypeIcon](type) {
         return this.jobTypeIcon[type] || 'help';
```

modules/Hoot/ui/managePanel/jobs/paging.js
```diff
@@ -21,7 +21,7 @@ export default class Paging {
     }
 
     getCurrentPage() {
-        return this.pageElement.text();
+        return JSON.parse( this.pageElement.text() );
```
brianhatchl: I wondered how this was working at all and looks like javascript will coerce the string to a number automatically in the comparison.  Try this in the browser console:
"2" > 1 = true
4 > "5" = false
"F" > 1 = false

If we want to be safe and coerce it to number in code I think `Number( this.pageElement.text() )` is a more meaningful code to future devs reading it.

modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -173,13 +173,26 @@ export default class Jobs extends Tab {
...
     async loadJobs() {
         let jobsRunning = await Hoot.api.getJobsRunning();
         let jobsHistory = await Hoot.api.getJobsHistory(this.params);
-        await Hoot.layers.refreshLayers();
+
+        let checkTotal = this.jobsHistoryTotalCheck(jobsHistory);
+
+        if ( jobsHistory.total > 0 && checkTotal ) {
+            await Hoot.layers.refreshLayers();
+        }
```
brianhatchl: Instead of all this, can we just do this?
`if ( jobsHistory.total > this.total ) {`

modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -364,10 +377,10 @@ export default class Jobs extends Tab {
 
     populateJobsHistory( jobs ) {
```
brianhatchl, line 368: When there are jobs present they aren't shown.  There is a line of code further down that needs to be adjusted now that we are passing the jobs object and not the jobs array.

brianhatchl (changes requested):

modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -176,10 +176,13 @@ export default class Jobs extends Tab {
     async loadJobs() {
         let jobsRunning = await Hoot.api.getJobsRunning();
         let jobsHistory = await Hoot.api.getJobsHistory(this.params);
-        await Hoot.layers.refreshLayers();
+
+        if ( jobsHistory.total > 0 && this.total ) {
```
brianhatchl: Don't we want to compare the jobs total in the current response with the existing jobs total and only call refresh layer if the total increases?  I think this will still call refresh layers when the total is 10 but has not changed.

brianhatchl (approved):

Issue #1754, opened 2020-03-23
Clip function doesn't seem to be working


ryanmcweeny:
When trying to use either of the clip dataset tools nothing occurs despite Hootenanny saying the task was completed. The dev console gave the following message.
<img width="556" alt="Screen Shot 2020-03-23 at 11 00 34 AM" src="https://user-images.githubusercontent.com/24612000/77331426-1566c500-6cf7-11ea-8378-023d706b4aa6.png">

brianhatchl:
Did you have a folder selected in the clip dialog (under Path)?
![image](https://user-images.githubusercontent.com/2600932/77331782-92923a00-6cf7-11ea-89d6-7e39e5e19a18.png)

ryanmcweeny:
I thought I had. The path usually has one selected automatically. However, I'm trying it now and everything is working with all vector files. I'm chocking this up to user error and marking as resolved.

ryanmcweeny:
So what actually happens is if you select the button to resolve all remaining issues in a merged layer and then try to make a clip of that layer the Path will no longer autopopulate or provide a dropdown menu with your folders.  This can be seen in the images below. This can be resolved by refreshing the page.
<img width="389" alt="Screen Shot 2020-03-23 at 1 31 31 PM" src="https://user-images.githubusercontent.com/24612000/77345459-05f17700-6d0b-11ea-94f4-5d983269b92d.png">
<img width="1003" alt="Screen Shot 2020-03-23 at 1 31 55 PM" src="https://user-images.githubusercontent.com/24612000/77345466-08ec6780-6d0b-11ea-86df-630b3f1108c8.png">

PR #1756, opened 2020-03-24
removing unnecessary removeLayer call, prevents folderId access w/in …

jackgrossman18:
ref #1754

### DIFF
--- a/modules/Hoot/tools/clipDataset.js
+++ b/modules/Hoot/tools/clipDataset.js
@@ -6,6 +6,7 @@
 
 import _forEach from 'lodash-es/forEach';
 import _map     from 'lodash-es/map';
+import _find    from 'lodash-es/find';
 
 import FormFactory from './formFactory';
 
@@ -93,7 +94,17 @@ export default class ClipDataset {
 
         let tableBody = table.append( 'tbody' );
 
+        let checkMerge = _find( loadedLayers, function(layer) {
+            if (layer.isMerged) {
+                return Hoot.layers.loadedLayers[ layer.id ] = layer;
+            }
+        } );
+
         _forEach( loadedLayers, layer => {
+            if ( checkMerge && checkMerge.id !== layer.id  ) {
+                return;
+            }
+
             let mapId = layer.id;
 
             tableBody
--- a/modules/Hoot/tools/clipDataset.js
+++ b/modules/Hoot/tools/clipDataset.js
@@ -6,7 +6,6 @@
 
 import _forEach from 'lodash-es/forEach';
 import _map     from 'lodash-es/map';
-import _find    from 'lodash-es/find';
 
 import FormFactory from './formFactory';
 
--- a/modules/Hoot/tools/clipDataset.js
+++ b/modules/Hoot/tools/clipDataset.js
@@ -94,16 +94,10 @@ export default class ClipDataset {
 
         let tableBody = table.append( 'tbody' );
 
-        let checkMerge = _find( loadedLayers, function(layer) {
-            if (layer.isMerged) {
-                return Hoot.layers.loadedLayers[ layer.id ] = layer;
-            }
-        } );
+        let loadedList = Object.values(loadedLayers);
+        let clipLayers = loadedList.some(l => l.isMerged) ? loadedList.filter(l => l.isMerged) : loadedList;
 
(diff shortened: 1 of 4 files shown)


modules/Hoot/ui/conflicts/resolve.js
```diff
@@ -3,8 +3,10 @@
  * Project: hootenanny-ui
  * @author Matt Putipong - matt.putipong@radiantsolutions.com on 5/24/18
  *******************************************************************************************************/
+import _forEach from 'lodash-es/forEach';
 
 import PublishBookmark from '../modals/publishBookmark';
+import { load } from 'js-yaml';
```
brianhatchl: is this used?

modules/Hoot/ui/conflicts/resolve.js
```diff
@@ -82,20 +84,20 @@ export default class Resolve {
...
+        let conflateController = this.sidebar.forms.conflate.controller;
+
+        let loadedLayers = Hoot.layers.loadedLayers;
+
+        _forEach( loadedLayers, lyr => {
+
+            if ( !lyr.isMerged ) {
+                Hoot.layers.removeLoadedLayer( lyr.id );
+            }
+        } );
```
brianhatchl: This fixes the bug, but introduces another.  By removing the source layers, the "show inputs" button is broken.
image

This makes me wonder what the point of removing the merged layer, only to add it back in might be.  Maybe we don't need to do that at all.

Maybe just forcing a reload of the layer data with review relations updated to resolved?  Would have to look at the http request that Resolve All Remaining makes, it might update the review relations on the back end, then maybe there is another method in iD to force a reload of the tiled layer data.

brianhatchl (changes requested):

jackgrossman18:
@brianhatchl I added a method to the sidebar controller class to hide the switch between the inputs of the merged layer, would this be suitable instead of forcing a reload of the layer data?

brianhatchl:
I think we want to keep it because it is still a merged layer and should act like if we added the merged layer from the add datasource button.

Maybe just try removing these two lines from the original:
```
        // update layer	
        Hoot.layers.removeLayer( layer.id );	
        Hoot.layers.loadLayer( key, true );	
```

I noticed other defects like when I go to resolve all remaining the count shown is zero when there are unresolved reviews.  Have you noticed this?

jackgrossman18:
@brianhatchl I added the proper count for the unresolved reviews, also updated clipDataset

modules/Hoot/ui/sidebar/sidebarController.js
```diff
@@ -182,6 +182,10 @@ class SidebarController {
             .text('delete_outline');
     }
 
+    hideShowLayersButton() {
+        d3.select('.showlayers').classed('hidden', true);
+    }
+
```
brianhatchl: no longer needed

modules/Hoot/ui/conflicts/resolve.js
```diff
@@ -82,20 +82,11 @@ export default class Resolve {
...
 
         // enter controller refresh state
         conflateController.text.html( 'Refreshing &#8230;' );
 
-        // update layer
-        Hoot.layers.removeLayer( layer.id );
-        Hoot.layers.loadLayer( key, true );
-
         // exit controller refresh state
         conflateController.text.html( layer.name );
```
brianhatchl: probably don't need these anymore

modules/Hoot/tools/clipDataset.js
```diff
@@ -93,7 +94,17 @@ export default class ClipDataset {
...
+            if (layer.isMerged) {
+                return Hoot.layers.loadedLayers[ layer.id ] = layer;
+            }
+        } );
+
         _forEach( loadedLayers, layer => {
+            if ( checkMerge && checkMerge.id !== layer.id  ) {
+                return;
+            }
+
```
brianhatchl: I find this kinda hard to follow.  Maybe something more "fluent" like:
```
        let loadedList = Object.values(loadedLayers);
        let clipLayers = loadedList.some(l => l.isMerged) ? loadedList.filter(l => l.isMerged) : loadedList;
        _forEach( clipLayers, layer => {

```

brianhatchl (changes requested):

brianhatchl (approved):

PR #1761, opened 2020-03-28
handle a missing mapid in the url hash

brianhatchl:
This fixes a silent failure (js error) when a mapid in the url hash for primary or secondary does not exist.
Now the UI will show a warning and remove the missing mapid from the url hash


brianhatchl:
You can test this by putting a non-existent map id in the url hash for `primary` or `reference`

bobi7 (approved):
💯

Issue #1765, opened 2020-04-06
Large import files don't start spinner until file is uploaded


brianhatchl:
This can make the UI seem like the import hasn't started yet and lead to multiple clicks to start the import.

PR #1767, opened 2020-04-09
Make large file upload requests cancelable

jackgrossman18:
ref #1765

### DIFF
--- a/modules/Hoot/managers/api.js
+++ b/modules/Hoot/managers/api.js
@@ -59,6 +59,7 @@ export default class API {
                 status     = response.status;
                 statusText = response.statusText;
                 type       = 'error';
+
             } else {
                 message = err.message;
                 status  = 500;
@@ -354,7 +355,6 @@ export default class API {
 
         let cancelURL = `${this.baseUrl}/ingest/ingest/upload`;
 
-
         axios.get( cancelURL, { cancelToken: cancelToken.token } )
         .catch(thrown => {
             if (axios.isCancel(thrown)) {
@@ -367,7 +367,7 @@ export default class API {
             Hoot.message.alert( alert );
             }
         });
-        cancelToken.cancel('Request canceled.');
+        cancelToken.cancel('Request cancelled.');
     }
 
     getJobsHistory( data ) {
--- a/modules/Hoot/managers/api.js
+++ b/modules/Hoot/managers/api.js
@@ -349,6 +349,27 @@ export default class API {
             .then( resp => resp.data );
     }
 
+    cancelUpload() {
+
+        let cancelURL = `${this.baseUrl}/ingest/ingest/upload`;
+
+        const source = axios.CancelToken.source();
+
+        axios.get( cancelURL, { cancelToken: source.token } )
+        .catch(thrown => {
+            if (axios.isCancel(thrown)) {
+            console.log(thrown.message);
+            } else {
+            let alert = {
+                message: 'Failed to cancel upload.',
+                type: 'warn'
+            };
+            Hoot.message.alert( alert );
+            }
+        });
+        source.cancel('Request canceled.');
+    }
+
     getJobsHistory( data ) {
         const params = {
             path: '/jobs/history',
--- a/modules/Hoot/managers/api.js
+++ b/modules/Hoot/managers/api.js
@@ -59,7 +59,6 @@ export default class API {
                 status     = response.status;
                 statusText = response.statusText;
                 type       = 'error';
-
             } else {
                 message = err.message;
                 status  = 500;
@@ -349,25 +348,6 @@ export default class API {
 
         return this.request( params )
             .then( resp => resp.data );
-    }
-
-    cancelUpload( cancelToken ) {
-
-        let cancelURL = `${this.baseUrl}/ingest/ingest/upload`;
-
-        axios.get( cancelURL, { cancelToken: cancelToken.token } )
-        .catch(thrown => {
-            if (axios.isCancel(thrown)) {
-            console.log(thrown.message);
-            } else {
-            let alert = {
-                message: 'Failed to cancel upload.',
-                type: 'warn'
-            };
-            Hoot.message.alert( alert );
-            }
-        });
-        cancelToken.cancel('Request cancelled.');
     }
 
     getJobsHistory( data ) {
@@ -772,10 +752,10 @@ export default class API {
                 INPUT_TYPE: data.INPUT_TYPE,
                 INPUT_NAME: data.INPUT_NAME,
                 NONE_TRANSLATION: data.NONE_TRANSLATION,
-                FOLDER_ID: data.folderId,
+                FOLDER_ID: data.folderId
             },
             data: data.formData,
-            cancelToken: cancelToken.token
+            cancelToken: cancelToken
         };
 
         if ( data.ADV_UPLOAD_OPTS && data.ADV_UPLOAD_OPTS.length ) {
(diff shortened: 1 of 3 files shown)


brianhatchl (changes requested):
I tried this but it doesn't abort the pending upload request which can take a long time if the upload file is large.  Here is the url of the request that's still pending:
http://localhost:8080/hoot-services/ingest/ingest/upload?TRANSLATION=NONE.js&INPUT_TYPE=OSM&INPUT_NAME=usa_texas&NONE_TRANSLATION=true&FOLDER_ID=58

Because the hoot api requests use axios, I think you'll need to figure out how to create and keep reference to a cancel token https://github.com/axios/axios#cancellation

modules/Hoot/managers/api.js
```diff
@@ -349,6 +349,27 @@ export default class API {
             .then( resp => resp.data );
     }
 
+    cancelUpload() {
+
+        let cancelURL = `${this.baseUrl}/ingest/ingest/upload`;
+
+        const source = axios.CancelToken.source();
+
+        axios.get( cancelURL, { cancelToken: source.token } )
```
brianhatchl: I don't think this will work.  You'll need to pass the cancel token to the original upload request, not issue a new request when cancel is called.  I think you'll want to pass the cancelToken to `Hoot.api.uploadDataset` in `importDataset.js`.  https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/Hoot/ui/modals/importDataset.js#L417

If you go back to the hoot2x branch version `source.cancel('Request canceled.');`  would be added here https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/Hoot/ui/modals/importDataset.js#L502

brianhatchl (changes requested):

modules/Hoot/managers/api.js
```diff
@@ -58,6 +59,7 @@ export default class API {
                 status     = response.status;
                 statusText = response.statusText;
                 type       = 'error';
+
```
brianhatchl: remove space to minimize diff

modules/Hoot/managers/api.js
```diff
@@ -349,6 +351,25 @@ export default class API {
...
+            let alert = {
+                message: 'Failed to cancel upload.',
+                type: 'warn'
+            };
+            Hoot.message.alert( alert );
+            }
+        });
+        cancelToken.cancel('Request cancelled.');
+    }
+
```
brianhatchl: We don't need a separate method in api.js and don't need to make another axios request.
Instead, just call 
`cancelToken.cancel('Request cancelled.');`
in import.js line 527.

modules/Hoot/managers/api.js
```diff
@@ -751,9 +772,10 @@ export default class API {
                 INPUT_TYPE: data.INPUT_TYPE,
                 INPUT_NAME: data.INPUT_NAME,
                 NONE_TRANSLATION: data.NONE_TRANSLATION,
-                FOLDER_ID: data.folderId
+                FOLDER_ID: data.folderId,
```
brianhatchl: no comma here

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -414,61 +418,78 @@ export default class ImportDataset {
...
-                    }
-
-                    this.submitButton
-                        .select( 'span' )
-                        .text( 'Import' );
-                } else {
-                    message = 'Import job complete';
+        this.loadingState();
+
+        this.processRequest =  Hoot.api.uploadDataset( data, this.source )
```
brianhatchl: Let's call uploadDataset with `this.source.token`

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -80,6 +81,9 @@ export default class ImportDataset {
         this.advOpts.forEach(opt => {
             opt.hidden = true;
         });
+
+        this.source = axios.CancelToken.source();
+
```
brianhatchl: move this to constructor()

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -414,61 +418,78 @@ export default class ImportDataset {
...
                 }
 
                 Hoot.message.alert( { message, type, keepOpen } );
+            }
+        } )
+        .finally( () => {
+            this.container.remove();
+            Hoot.events.emit( 'modal-closed' );
+        } );
+
```
brianhatchl, line 484: remove to minimize diff

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -487,6 +508,7 @@ export default class ImportDataset {
     }
 
     loadingState() {
+
```
brianhatchl: remove to minimize diff

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -414,61 +418,78 @@ export default class ImportDataset {
...
+        .then( async resp => {
+            let message;
+
+            if ( resp.message === 'Request canceled'  ) {
+                Hoot.message.alert( {
+                    message: resp.message,
+                    type: 'warn'
+                } );
+            }
+
```
brianhatchl: I'm not sure we need this both here and below.  Seems like we call `source.cancel()` with "Request cancelled" with two Ls so not sure this ever gets hit.

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -499,7 +521,10 @@ export default class ImportDataset {
 
         // overwrite the submit click action with a cancel action
         this.submitButton.on( 'click', () => {
-            Hoot.api.cancelJob(this.jobId);
+            if ( this.jobId ) {
+                Hoot.api.cancelJob(this.jobId);
+            }
+            Hoot.api.cancelUpload( this.source );
```
brianhatchl: Did we find out if calling this on a completed request throws an error?
If so we could probably put it in an `} else { }` with the `if`

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -414,61 +418,78 @@ export default class ImportDataset {
...
+        .catch( err => {
+            console.error(err);
+
+            if ( err.message === 'Request cancelled.' ) {
                 Hoot.message.alert( {
-                    data: resp.data,
-                    message: message,
-                    status: 200,
-                    type: resp.type
+                    message: 'Upload request cancelled',
```
brianhatchl: might as well use `err.message` here and make it match what we want to show the user.

modules/Hoot/managers/api.js
```diff
@@ -751,9 +772,10 @@ export default class API {
...
             },
-            data: data.formData
+            data: data.formData,
+            cancelToken: cancelToken.token
```
brianhatchl: Let's make it so we pass the cancelToken itself (see comment below)

brianhatchl (changes requested):
It's working :tada: 
But I have a few implementation/style changes for you to make.

modules/Hoot/ui/modals/importDataset.js
jackgrossman18: this prevents the cancelJob from running without any jobId. I've pushed all the changes you requested. This change will prevent cancelJob from being run if the upload is cancelled prior to the first jobId response. Once that happens, you wont' be able to cancel the upload

brianhatchl: Right, but what happens if the upload request has completed, there is now a jobId, and the cancel button is clicked.  The hoot cancelJob() will be called but so will cancelUpload() which may throw an error.  Putting cancelUpload in an else here will make it either one or the other.

brianhatchl (approved):

PR #1771, opened 2020-04-23
Fixed #1752. added group jobs action to jobs panel

### LINKED ISSUE #1752: Add some kind of pointer for derive changeset jobs back to source in Output column
For instance this might be the conflation project # and task # (for create differential changeset)
Or it might be the input dataset names

bobi7:
when user selected to see a jobs 'family' it will show each job related to the one selected for ease of seeing a workflow run

ref #1752


brianhatchl (approved):
:sunglasses:

PR #1777, opened 2020-04-30
Fixed #1776. user 'last authorized' shows up as undefined for > year

### LINKED ISSUE #1776: admin panel 'last authorized' shows up as undefined
Screen Shot 2020-04-30 at 9 08 05 AM

bobi7:
ref #1776


brianhatchl:
Can you add a test for durations over a year to `test/hoot/spec/utilities.js`?

This might be the best fix, but seems like there's an improvement in the logic that could be made to return the actual number of years ago.

brianhatchl (approved):
:doughnut: :coffee:

PR #1779, opened 2020-05-05
URL values in raw tag editor override taginfo link with open in browser link

jackgrossman18:
ref #1769

### DIFF
--- a/modules/ui/raw_tag_editor.js
+++ b/modules/ui/raw_tag_editor.js
@@ -142,6 +142,7 @@ export function uiRawTagEditor(context) {
             .append('input')
             .property('type', 'text')
             .attr('class', 'value')
+            .attr('xlink:href', function(d) { return '#' + d.value; })
             .attr('maxlength', 255)
             .call(utilNoAuto)
             .on('blur', valueChange)
@@ -177,14 +178,22 @@ export function uiRawTagEditor(context) {
 
                 var reference;
 
-                if (isNaN( d.value ) && typeof d.value === 'string' && d.value.indexOf('http') === 0) {
-                reference = uiTagLinkUrl({ url: d.value });
-                }
-
                 // Hoot: Override tag reference with tag copy
                 var select = uiTagSelectCopy(context);
 
-                row.call( select );
+                if (isNaN(d.value) && typeof d.value === 'string' && d.value.indexOf('http') === 0) {
+                    reference = uiTagLinkUrl({ url: d.value });
+
+                row.select('.inner-wrap')      // propagate bound data
+                .call(reference.button);
+
+                row.call(reference.body);
+
+                row.select('button.remove');   // propagate bound data
+
+                } else {
+                    row.call( select );
+                }
             });
 
         items.selectAll('input.key')
--- a/modules/ui/raw_tag_editor.js
+++ b/modules/ui/raw_tag_editor.js
@@ -16,13 +16,12 @@ import { services } from '../services';
 import { svgIcon } from '../svg';
 import { uiDisclosure } from './disclosure';
 import { uiTagSelectCopy } from './tag_select_copy';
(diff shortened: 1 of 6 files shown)


modules/ui/tag_link_url.js
```diff
@@ -0,0 +1,77 @@
...
+            .data([0]);
+
+        _button = _button.enter()
+            .append('button')
+            .attr('class', 'tag-reference-button')
+            .attr('tabindex', -1)
+            .merge(_button);
+
+        _button
+            .attr('title', t('icons.open_link'))
```
brianhatchl, line 39: We need to add the value for `icons.open_link` to core.yaml

modules/ui/raw_tag_editor.js
```diff
@@ -175,10 +174,24 @@ export function uiRawTagEditor(context) {
...
-                row.call( select );
+                if (isNaN(d.value) && typeof d.value === 'string' && d.value.indexOf('http') === 0) {
+                    reference = uiTagLinkUrl({ url: d.value });
+
+                row.select('.inner-wrap')      // propagate bound data
+                .call(reference.button);
+
+                row.call(reference.body);
+
+                row.select('button.remove');   // propagate bound data
```
brianhatchl, line 200: Not sure if we need this or can just call
`row.call(reference)`

the difference in syntax might be where this was plucked from iD `v2.14.`3 and hoot-ui is only on iD `v2.12.1`

modules/ui/raw_tag_editor.js
```diff
@@ -339,6 +352,14 @@ export function uiRawTagEditor(context) {
                 _newRow = undefined;
             }
 
+            //Enable tag link url
+            if (this.value.indexOf('http') === 0) {
+                var reference = uiTagLinkUrl({ url: this.value });
+                var row = d3_select(this.parentNode.parentNode.parentNode);
+                row.select('.inner-wrap').call(reference.button);
+                row.call(reference.body);
+            }
```
brianhatchl: When a new tag is added with a url value, this code is invoked, but it doesn't remove or update the existing tag copy button, leading to the ui appear like this:

Screenshot from 2020-05-19 17-33-57

When an existing feature with a url tag is selected (or another feature is selected and this feature then re-selected) the tag table renders fine.

brianhatchl (changes requested):

modules/ui/raw_tag_editor.js
jackgrossman18: in testing removing body from line 188, it results in empty rows

remove_body

modules/ui/raw_tag_editor.js
```diff
@@ -339,6 +363,14 @@ export function uiRawTagEditor(context) {
...
             }
 
+            // //Enable tag link url
+            // if (this.value.indexOf('http') === 0) {
+            //     var reference = uiTagLinkUrl({ url: this.value });
+            //     var row = d3_select(this.parentNode.parentNode.parentNode);
+            //     row.select('.inner-wrap').call(reference.button);
+            //     row.call(reference.body);
+            // }
+
```
brianhatchl: remove if not needed

modules/ui/raw_tag_editor.js
```diff
@@ -175,10 +175,34 @@ export function uiRawTagEditor(context) {
                     bindTypeahead(key, value);
                 }
 
+                var isRelation = (_entityID && context.entity(_entityID).type === 'relation');
+                var reference;
+
                 // Hoot: Override tag reference with tag copy
                 var select = uiTagSelectCopy(context);
 
                 row.call( select );
```
brianhatchl: Should call this after other button types or move checkmark above column

image

brianhatchl:
The border radius needs to be moved to the tag copy button
image

brianhatchl (approved):

PR #1783, opened 2020-06-14
fixed #1778. tasking manager panel added to hoot ui

### LINKED ISSUE #1778: add panel allowing user to run tasking manager tasks
Hoot would pull all cells for a TM campaign and run the cells in series

show status of each cell incase error occurs

bobi7:
ref #1778


modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -829,7 +832,7 @@ export default class Jobs extends Tab {
                                             deriveType : 'Merged changeset'
                                         };
 
-                                        Hoot.api.deriveChangeset( data, params )
+                                        Hoot.api.createDifferentialChangeset( data, params )
```
brianhatchl: I don't understand why this change was made, `deriveChangeset` is a more general term than `createDifferentialChangeset` which implies a specific type of conflation (Differential) was used

modules/Hoot/managers/api.js
```diff
@@ -65,7 +67,8 @@ export default class API {
                 type    = 'error';
             }
 
-            if ( status === 401 && statusText === 'Unauthorized' ) {
+            // only redirect to login page for unauthorized hoot services requests
+            if ( status === 401 && statusText === 'Unauthorized' && request.url.includes('/hoot-services') ) {
```
brianhatchl, line 75: Add more comment about how this object also makes requests to TM if that's the other 401 response that needs separate handling from Hoot login redirect.  If that's the case, what is the handling of TM 401s?

brianhatchl: Looking at the rest of the urls in api.js, how is the `/hoot-services` path inferred?  So how does `/jobs` get directed to `hoot-services/jobs`.

modules/Hoot/ui/modals/grailDatasetPicker.js
```diff
@@ -83,7 +83,7 @@ export default class GrailDatasetPicker {
 
         this.params.replacement = true;
 
-        Hoot.api.deriveChangeset( data, this.params )
+        Hoot.api.createDifferentialChangeset( data, this.params )
```
brianhatchl: Comment about this in https://github.com/ngageoint/hootenanny/pull/4095

brianhatchl (changes requested):
Some initial feedback just from code diff.  Will update and run next.

modules/Hoot/managers/api.js
bobi7: the path is inferred because the params object will have path if services, url if other

bobi7: For TM requests that return 401s we will log an error message still and not load that panel but that shouldnt log the user out of hootenanny

modules/Hoot/ui/managePanel/jobs.js
bobi7: you're right

modules/Hoot/managers/api.js
brianhatchl: I see, it uses baseUrl which is https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/Hoot/config/apiConfig.js#L22

bobi7: correct. i added a comment for future clarity

https://github.com/ngageoint/hootenanny-ui/pull/1783/commits/d75a853f46c02d555daa7cbd07628d4328f7864d#diff-700cc137e7a64a224df0acdfab9d4aacR40-R45

modules/Hoot/ui/managePanel/TaskingManagerPanel.js
```diff
@@ -0,0 +1,373 @@
...
+        super( instance );
+
+        this.name = 'Tasking Manager';
+        this.id   = 'taskingManagerPanel';
+        this.currentProject = null;
+
+        this.deriveTypeOpts = {
+            name: 'deriveType',
+            readonly: 'readonly',
+            options: [ 'Adds only', 'Cut & Replace', 'Differential changeset' ]
```
brianhatchl: Let's remove "changeset" from the last option.  Thinking on these terms "Adds only" might be confused with Differential.  Might need to think about these more.  Or have tool tips to show details of what they mean.

modules/Hoot/managers/api.js
```diff
@@ -1540,4 +1529,127 @@ export default class API {
...
+                };
+
+                Hoot.message.alert( alert );
+            } );
+    }
+
+    /**
+     * Retrieves all the tasks from tasking manager for the specified project
+     */
+    getTMTasks( projectId ) {
```
brianhatchl, line 1560: Do you think we should poll this when the TM tab is active?  I noticed that the TM UI will refresh when Hoot locks a task so maybe Hoot should refresh if a user locks a task in TM?

modules/Hoot/ui/managePanel/TaskingManagerPanel.js
```diff
@@ -0,0 +1,373 @@
...
+            .classed( 'taskingManager-details', true );
+        details.append( 'label' )
+            .text( 'Created At: ' );
+        details.append( 'span' )
+            .text( project => {
+                let { created, author } = project.properties;
+                created = new Date( created ).toLocaleString();
+
+                return `${ created } by ${ author }`;
+            } );
```
brianhatchl: Let's make this look more like TM with the project # in the title, no label for `Description:` and a matching layout for created by and created on.
image

No need for the last updated, priority, map or percent complete.

modules/Hoot/ui/managePanel/jobs.js
```diff
@@ -49,6 +49,9 @@ export default class Jobs extends Tab {
             delete: 'delete',
             derive_changeset: 'change_history',
             upload_changeset: 'cloud_upload',
+            bulk_add: 'add_to_photos',
+            bulk_replace: 'flip_to_front',
+            bulk_differential: 'change_history',
```
brianhatchl, line 54: I haven't gotten my dev envi setup to actually run the bulk tasks yet, can you include a screenshot of what these job entries and icons look like?

css/hoot/components/managePanel/_taskingManager.scss
```diff
@@ -0,0 +1,167 @@
...
+        &[status="Running"] {
+            border: 4px solid #ffdca3;
+
+            .taskingManager-action-buttons {
+                margin-right: 10px;
+            }
+        }
+
+        &[status=""] {
+            background-color: #cdcdcd;
```
brianhatchl: maybe a bit lighter gray?  did we even consider white?

modules/Hoot/ui/managePanel/jobs.js
bobi7: image

Uses same icons as the single, non-bulk, jobs

css/hoot/components/managePanel/_taskingManager.scss
bobi7: I like the lighter gray more because in tasking manager the fill is a light grey when the task is not locked and hasn't been worked on

modules/Hoot/managers/api.js
bobi7: The 1 problem I see right now is that it would ruin the 'Running' status since that is something I added and it isn't a part of TM of course

modules/Hoot/ui/managePanel/TaskingManagerPanel.js
```diff
@@ -0,0 +1,375 @@
...
+        let description = body.append( 'div' )
+            .classed( 'taskingManager-description', true );
+        description.append( 'span' )
+            .text( project => project.properties.short_description );
+
+        let details = body.append( 'div' )
+            .classed( 'taskingManager-details', true );
+        details.append( 'span' ).text( project => {
+            let { created, author } = project.properties;
+            created = new Date( created ).toLocaleString();
```
brianhatchl: nitpick: No time component here, just date

modules/Hoot/ui/managePanel/TaskingManagerPanel.js
```diff
@@ -0,0 +1,375 @@
...
+        runAllBtn.property( 'disabled', false );
+    }
+
+    async loadTaskTable( project ) {
+        const tmPanel = this;
+        this.tasksTable.selectAll( '.taskingManager-item' ).remove();
+        this.currentProject = project;
+
+        this.tasksContainer.select( '.taskHeader-title' )
+            .text( `Tasks for: ${ this.currentProject.properties.name }`);
```
brianhatchl: Just the project # and then name here, no "Tasks for:"

brianhatchl (changes requested):
couple more ui nitpicks

brianhatchl (approved):

PR #1786, opened 2020-07-12
Fixed #1784. Add user customizable POI conflation search radii to advanced options

### LINKED ISSUE #1784: Add user customizable POI conflation search radii to advanced options
Working with @curranMapper, we've recognized some conflation scenarios where hoot's default configured search radius distances for POI conflation aren't adequate. It would be useful to give users the option to customize these distances as needed.

As part of [these changes](https://github.com/ngageoint/hootenanny/pull/4096), hoot now accepts incoming POI feature search radii distances as JSON, whereas before they weren't customizable unless you were running from the command line and made edits to `rules/Poi.js`. Now, we'd like to expose this functionality via the UI. Definitely open to other ideas, but I'm thinking that under the "Poi" section of the adv opts, add a dialog that shows each feature type hoot POI conflation supports along with their default search radii. Then allow the user to change those distances from there.

* The default POI search radii are defined in [this file](https://github.com/ngageoint/hootenanny/blob/master/conf/core/PoiSearchRadii.json), which is configured with the `poi.search.radii.config` config option. So, this file can be read in to get the defaults.
* Another config option, `poi.search.radii` has been set up to override the aforementioned file and pass in the search radii as JSON from the UI.
* The JSON format is very simple. 'key' and 'distance' are required and 'value' is optional. If 'value' is left out, then all features with the tag key will be matched with the same search radius.
* **BONUS IDEA:** Additionally if we want, we could allow users to add search radii for additional types that aren't in the hoot defaults. The one caveat is that if hoot doesn't have the user added type defined in its schema then it won't able to conflate it as a POI (there are a lot of POI types already in the schema, though). In those cases the point *should* get conflated with generic point conflation, which is a little more stripped down in functionality and then will ignore the specified POI search radius and use the custom radius its configured with. There are ways to query hoot's schema to know what types are considered POI's, but I don't think that information is available outside of the core. So, if we did this we may want to add the capability to query for that information first. Will leave the decision of whether we want to implement this particular functionality up to Tradecraft.

bobi7:
ref #1784

### DIFF
--- a/modules/Hoot/ui/sidebar/advancedOpts.js
+++ b/modules/Hoot/ui/sidebar/advancedOpts.js
@@ -762,30 +762,54 @@ export default class AdvancedOpts {
         }
 
         let options = { advanced: {}, cleaning: [] };
+        let jsonFieldIds = new Set();
         let conflateType = d3.select( '#conflateType' ).property( 'value' ).toLowerCase();
-        this.contentDiv.selectAll( '.form-group' ).each( function(d) {
+        this.contentDiv.selectAll( '.form-group' ).each( function( formGroup ) {
             let selection = d3.select( this );
-            let isCleaning = d.name === 'Cleaning';
+            let isCleaning = formGroup.name === 'Cleaning';
 
-            selection.selectAll( '.hoot-form-field' ).each( function(d) {
+            selection.selectAll( '.hoot-form-field' ).each( function( formField ) {
 
-                if ( !d.send ) {
+                if ( !formField.send ) {
                     return; // if no d.send, then input value never changed from default...
                 }
 
                 const value = d3.select( this ).select( 'input' )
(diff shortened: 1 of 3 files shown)


modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -762,30 +762,54 @@ export default class AdvancedOpts {
...
                 } else {
-                    options.cleaning.push( d.id );
+                    options.cleaning.push( formField.id );
                 }
             });
+
+            for ( let id of jsonFieldIds ) {
+                options.advanced[ id ] = JSON.stringify( options.advanced[ id ] );
+                jsonFieldIds.delete( id );
+            }
```
bobi7, line 812: need to stringify outside the loop because never know if more data will be added to the object

bobi7:
looks good

brianhatchl (approved):

PR #1787, opened 2020-07-14
fixed test due to newly made format for tags params to be object and …

bobi7:
…not string


brianhatchl:
good catch, fixed myself in https://github.com/ngageoint/hootenanny-ui/commit/9b336f46b9a2f025639b8b2cdd5831da93131bf8 before I saw this.

PR #1800, opened 2020-08-03
deduplicate layer names on multi-dataset import

### LINKED ISSUE #1799: Import Multiple overwrites existing datasets with same name
Instead it should append ` (1)` to avoid conflict

brianhatchl:
fixes #1799


brianhatchl:
Multiple filename and "as Single" names should get deconflicted (adding a `(#)` to the end).

image

The "as Single" name gets updated on text box blur.

jackgrossman18 (approved):
👍

PR #1803, opened 2020-08-14
Fix js error with translated tags table

### LINKED ISSUE #1802: Tag Schemas are not populating in All Tags section when changed
If you take a feature such as a railway in OSM tag schema and try and translate it to TDSv4 or anything besides the OSM the All Tags section will not populate and Hoot will inform you the feature is out of spec. No error seems to show up in the console. This was tried with several types of way and point features. using the MassNh_OSM_small test file.

brianhatchl:
fixes #1802

### DIFF
--- a/modules/Hoot/managers/translationManager.js
+++ b/modules/Hoot/managers/translationManager.js
@@ -220,4 +220,9 @@ export default class TranslationManager {
         }
         return values;
     }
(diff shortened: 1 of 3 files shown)


bobi7 (approved):
lgtm 🍎

PR #1809, opened 2020-09-14
add tagging for ec2 instances started by Jenkins jobs

gvoltz:
task in association to jira ticket: https://radiantsolutions.atlassian.net/browse/VGI-1659

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,15 @@ pipeline {
     parameters {
         booleanParam(name: 'UI', defaultValue: true)
         string(name: 'Box', defaultValue: 'default', description: 'Vagrant Box')
+        string(name: 'HOOT_UI_AWS_ENV_TAG', defaultValue: 'testing')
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -5,6 +5,15 @@ pipeline {
     parameters {
         booleanParam(name: 'UI', defaultValue: true)
         string(name: 'Box', defaultValue: 'default', description: 'Vagrant Box')
+        string(name: 'HOOT_UI_AWS_ENV_TAG', defaultValue: 'testing')
```
sputnam, line 8: @gvoltz Same comment as with hootenanny.  Do we want to make these configurable parameters?

gvoltz: It was more of a clutter vs. hard-coding in my mind for the tags in the Jenkinsfile. The params I'd bet would rarely be changed, but without the params the tags would be hard-coded from the environment namespace which I wanted to avoid.

sputnam (approved):

PR #1815, opened 2020-09-28
Fixed #1801. fixed where tm pull would use custom extent if not selected

### LINKED ISSUE #1801: TM pulls not always going to correct project folder
The folder gets its name from the comment url hash which is then kept in local storage.  I wonder if there was a failure to get the new value or somehow the pull was done without a launch from TM, but an old local storage value was still present.

bobi7:
ref #1801

### DIFF
--- a/modules/Hoot/tools/grailPull.js
+++ b/modules/Hoot/tools/grailPull.js
@@ -201,16 +201,6 @@ export default class GrailPull {
         } );
     }
 
-    // Checks if there is tm project and task info set. Also check that the custom
-    static customDataExtentEnabled() {
-        if ( sessionStorage.getItem('tm:project') && sessionStorage.getItem('tm:task') &&
-            Hoot.context.layers().layer('data').enabled() && Hoot.context.layers().layer('data').hasData() ) {
-            return true;
-        } else {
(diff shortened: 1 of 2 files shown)


modules/Hoot/tools/grailPull.js
```diff
@@ -200,6 +201,16 @@ export default class GrailPull {
         } );
     }
 
+    // Checks if there is tm project and task info set. Also check that the custom
+    static customDataExtentEnabled() {
+        if ( sessionStorage.getItem('tm:project') && sessionStorage.getItem('tm:task') &&
+            Hoot.context.layers().layer('data').enabled() && Hoot.context.layers().layer('data').hasData() ) {
```
brianhatchl: This line is checking the presence of gpx task bounds data?

brianhatchl (approved):
LGTM :martial_arts_uniform:

modules/Hoot/tools/grailPull.js
bobi7: yea and that its enabled just incase someone has a bounds but they dont want to use it for a pull, they can still disabled the bounds

PR #1820, opened 2020-10-12
Fixed #1819. handle cases where cut and replace returns no options

### LINKED ISSUE #1819: fix display in UI for when /info/getChangesetOptions returns empty json
If cut and replace doesnt return any options we want to handle errors from that value being null

in the tasking manager panel we will no longer show the options dialog box when the user selects 'cut & replace' when there aren't any options present

bobi7:
ref #1819


brianhatchl (approved):
:ok_hand: :us:

PR #1822, opened 2020-10-19
Fixed #1821.set conflation algorithm null when roads option is unchecked

### LINKED ISSUE #1821: dont send conflation algorithm when roads option is unchecked
https://radiantsolutions.atlassian.net/browse/VGI-1895

bobi7:
ref #1821


brianhatchl (approved):
:turtle: :turtle:

PR #1825, opened 2020-10-26
Fixed #1818. will auto select dataset with matching uuid for replacement

### LINKED ISSUE #1818: grail data picker to match on uuid snip if present
try to suggest dataset with matching uuid when running derive changeset replacement

bobi7:
ref #1818

### DIFF
--- a/modules/Hoot/ui/modals/grailDatasetPicker.js
+++ b/modules/Hoot/ui/modals/grailDatasetPicker.js
@@ -79,7 +79,6 @@ export default class GrailDatasetPicker {
 
     updateSubmitButton() {
         this.submitButton.attr( 'disabled', function() {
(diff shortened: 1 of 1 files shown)


modules/Hoot/ui/modals/grailDatasetPicker.js
```diff
@@ -57,8 +60,26 @@ export default class GrailDatasetPicker {
...
+                }
+            } );
+        }
+
+        return secondaryDataset;
+    }
+
     updateSubmitButton() {
         this.submitButton.attr( 'disabled', function() {
+            console.log(d3.select('#refDataset'));
```
brianhatchl: can remove?

brianhatchl (approved):
remove console log and you're G2G :haircut_man:

PR #1827, opened 2020-10-28
Fixed #1826. fixed delete stale data dropdown cutoff. user list scroll

### LINKED ISSUE #1826: make admin user table scrollable
This is cleaner looking but also helps with the issue of the 'Delete Stale Map Datasets' dropdown list getting cut off (tim noticed this when he wanted to clear up disk space) when a lot of users are present and its at the bottom of the page

bobi7:
ref #1826

### DIFF
--- a/css/hoot/components/managePanel/_admin.scss
+++ b/css/hoot/components/managePanel/_admin.scss
@@ -13,7 +13,7 @@
 
     tbody {
         display: block;
-        max-height: 600px;
+        max-height: 800px;
         overflow-y: auto;
     }
 
--- a/css/hoot/components/managePanel/_admin.scss
+++ b/css/hoot/components/managePanel/_admin.scss
@@ -6,8 +6,25 @@
     box-shadow: 0 3px 10px $dark-opacity10;
 
     thead {
+        display: block;
         color: $light;
         background: $dark1;
+    }
+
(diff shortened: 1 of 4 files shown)


brianhatchl:
Seems like an alternate fix could use this https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/lib/d3.combobox.js#L274-L286

I don't know why we are maintaining a hoot version of the d3.combobox.  It seems pretty identical except for a min item of 1 vs 2 and disabling the autocomplete code.  I don't know if git history reveals any rationale behind this choice, but utilizing the default d3.combobox seems cleaner.

brianhatchl:
Looking at history, we seem to have added an alternate `_value` from what is displayed as well as a `selectedID`.  Not sure if that warrants a hoot copy of the class, but what do you think about just adding the code to make the drop down go up?

bobi7:
Will look into that. I didn't see this code block and it also wasnt working properly so i guess i need to debug to see why the menu towards top wasnt triggered

brianhatchl:
The hoot version of combobox seems to have deleted the open towards top code, not sure why, but you could add it back in

brianhatchl (approved):
:jack_o_lantern: :chocolate_bar:

PR #1834, opened 2020-11-09
Fixed #1833. We can now split GeometryCollection into multiple features

### LINKED ISSUE #1833: export alpha shape download fails for GeometryCollection
UI doesn't have code to handle feature type GeometryCollection

bobi7:
ref #1833

### DIFF
--- a/modules/svg/data.js
+++ b/modules/svg/data.js
@@ -1,8 +1,9 @@
+import _cloneDeep from 'lodash-es/cloneDeep';
 import _flatten from 'lodash-es/flatten';
 import _isEmpty from 'lodash-es/isEmpty';
 import _reduce from 'lodash-es/reduce';
-import _union from 'lodash-es/union';
 import _throttle from 'lodash-es/throttle';
+import _union from 'lodash-es/union';
 
 import {
     geoBounds as d3_geoBounds,
@@ -121,6 +122,11 @@ export function svgData(projection, context, dispatch) {
         if (!gj) return null;
 
         if (gj.type === 'FeatureCollection') {
+            // Split geometry collection into seperate features
+            if (containsGeometryCollection(gj)) {
+                geometryCollectionToFeatures(gj);
+            }
+
             for (var i = 0; i < gj.features.length; i++) {
                 ensureFeatureID(gj.features[i]);
             }
@@ -158,6 +164,17 @@ export function svgData(projection, context, dispatch) {
 
     function isPolygon(d) {
         return d.geometry.type === 'Polygon' || d.geometry.type === 'MultiPolygon';
+    }
+
+
+    function containsGeometryCollection(gj) {
+        var features = getFeatures(gj);
+
+        var geomCollections = features.filter(function(feature) {
+            return feature.geometry.type === 'GeometryCollection';
+        });
+
+        return geomCollections.length > 0;
     }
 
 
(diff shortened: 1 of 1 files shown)


modules/svg/data.js
```diff
@@ -121,6 +122,11 @@ export function svgData(projection, context, dispatch) {
         if (!gj) return null;
 
         if (gj.type === 'FeatureCollection') {
+            // Split geometry collection into seperate features
+            if (containsGeometryCollection(gj)) {
+                geometryCollectionToFeatures(gj);
+            }
```
brianhatchl: I think GeometryCollection could also appear as a single feature, outside of a FeatureCollection.  It should be possible to support that too by making `geometryCollectionToFeatures` more like `ensureFeatureID` in that it takes a single feature at a time and operates on it.

brianhatchl (changes requested):
This is good and works, but...

modules/svg/data.js
```diff
@@ -161,6 +167,17 @@ export function svgData(projection, context, dispatch) {
     }
 
 
+    function containsGeometryCollection(gj) {
+        var features = getFeatures(gj);
+
+        var geomCollections = features.filter(function(feature) {
+            return feature.geometry.type === 'GeometryCollection';
+        });
```
brianhatchl: I think this could be made more performant by using `.some()` so that it can stop looping at the first GeometryCollection.

modules/svg/data.js
```diff
@@ -427,6 +444,45 @@ export function svgData(projection, context, dispatch) {
...
+
+        var newFeaturesList = _reduce(currentFeatures, function(featuresAggregate, featureItem) {
+            var features = [];
+
+            if (featureItem.geometry.type === 'GeometryCollection') {
+                var geometries = featureItem.geometry.geometries;
+
+                geometries.forEach( function(geom, index) {
+                    var newFeature = _cloneDeep(featureItem);
+                    newFeature.roles = newFeature.roles.split(';')[index];
```
brianhatchl: So `roles` isn't really a valid geojson property and neither is `relation-type`.  Not sure we need to handle it explicitly.  Rather than a deep clone maybe just copy the `properties` property to all output features.

brianhatchl (commented):
I like the approach of this refactoring, but a GeometryCollection can also have Points or LineStrings or Polygons as long as they are all the same geometry type.

modules/svg/data.js
```diff
@@ -157,7 +157,7 @@ export function svgData(projection, context, dispatch) {
 
 
     function isPolygon(d) {
-        return d.geometry.type === 'Polygon' || d.geometry.type === 'MultiPolygon';
+        return d.geometry.type === 'Polygon' || d.geometry.type === 'MultiPolygon' || d.geometry.type === 'GeometryCollection';
```
brianhatchl: This will assume all GeometryCollections should be styled like areas (with the inner shading).  Might I suggest a boolean expression like this to test the types in the geometries property:
```
|| (d.geometry.type === 'GeometryCollection' && d.geometry.geometries.every(function(g) {
+            return g.type === 'Polygon' || g.type === 'MultiPolygon';
+        }))
```

modules/svg/data.js
```diff
@@ -500,29 +500,41 @@ export function svgData(projection, context, dispatch) {
...
-                case 'MultiLineString':
-                    c = _flatten(c);
-                    break;
+            var c = [];
+            if (feature.geometry.geometries) {
+                feature.geometry.geometries.forEach( function(geom) {
+                    c = c.concat(flattenGeoms(geom.coordinates, geom.type));
+                });
+            } else {
+                c = flattenGeoms(feature.geometry.coordinates, feature.geometry.type);
```
brianhatchl, line 538: really good approach, better than turning geometry collection into feature array in the geojson

modules/svg/data.js
```diff
@@ -500,29 +500,41 @@ export function svgData(projection, context, dispatch) {
...
+    function flattenGeoms(c, type) {
+        /* eslint-disable no-fallthrough */
+        switch (type) {
+            case 'Point':
+                c = [c];
+            case 'MultiPoint':
+            case 'LineString':
+                break;
+
+            case 'GeometryCollection':
```
brianhatchl: I think this case line can be deleted, as the code will now pass geometries.

modules/svg/data.js
brianhatchl: LineStrings will appear like this:
image
vs
image

brianhatchl (changes requested):
I think you nailed it :hammer: 

couple finishing suggestions

brianhatchl (approved):

PR #1835, opened 2020-11-12
Fixed #1791. fix review bookmark reload

### LINKED ISSUE #1791: Load review bookmark shouldn't reload merged layer if present
But the code here to jump to a review depends on the whole merged layer loaded code path.

https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/Hoot/ui/managePanel/reviewBookmarks/reviewBookmarkNotes.js#L155-L167

So not a one or two line change, hence this new ticket.

bobi7:
ref #1791

### DIFF
--- a/modules/Hoot/ui/conflicts/info.js
+++ b/modules/Hoot/ui/conflicts/info.js
@@ -29,7 +29,7 @@ export default class ConflictMetadata {
         ];
 
         Hoot.context.history().on('change.reviewtagtable',
-            _debounce( () => this.buildTagTable.bind(this)(), 300)
+            _debounce( () => this.buildTagTable.bind(this), 300)
         );
     }
 
--- a/modules/Hoot/ui/managePanel/reviewBookmarks/reviewBookmarkNotes.js
+++ b/modules/Hoot/ui/managePanel/reviewBookmarks/reviewBookmarkNotes.js
@@ -150,17 +150,24 @@ export default class ReviewBookmarkNotes extends Tab {
      *   and the value in _forcedReviewableItem gets used to display review item.
      **/
     async jumpToReviewItem() {
-        Hoot.layers.removeAllLoadedLayers();
-        Hoot.ui.navbar.toggleManagePanel();
-
-        Hoot.ui.conflicts.data.forcedReviewItem = this.currentReviewable;
-
         let params = {
             name: this.bookmark.layerName,
             id: this.bookmark.mapId
         };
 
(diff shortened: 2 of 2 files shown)


modules/Hoot/ui/conflicts/info.js
```diff
@@ -29,7 +29,7 @@ export default class ConflictMetadata {
         ];
 
         Hoot.context.history().on('change.reviewtagtable',
-            _debounce( () => this.buildTagTable.bind(this)(), 300)
+            _debounce( () => this.buildTagTable.bind(this), 300)
```
bobi7, line 32: I didn't notice other places with the extra parenthesis so I removed them and fixed this issue i was seeing with tags table being duplicated.

The way i was able to reproduce the error was:

- select a review bookmark to load review
- remove dataset
- go back to same bookmark and load it
- should be 2 of the same table stacked on each other

modules/Hoot/ui/managePanel/reviewBookmarks/reviewBookmarkNotes.js
```diff
@@ -150,17 +150,24 @@ export default class ReviewBookmarkNotes extends Tab {
...
+        if ( Hoot.layers.loadedLayers[ params.id ] ) {
+            Hoot.ui.navbar.toggleManagePanel();
+            Hoot.ui.conflicts.data.forcedReviewItem = this.currentReviewable;
+
+            Hoot.ui.conflicts.graphSync.getRelationMembers( this.currentReviewable.relationId )
+                .then( members => Hoot.ui.conflicts.map.highlightLayer( members[ 0 ], members[ 1 ], true ) );
+        } else {
+            Hoot.layers.removeAllLoadedLayers();
+            Hoot.ui.navbar.toggleManagePanel();
+            Hoot.ui.conflicts.data.forcedReviewItem = this.currentReviewable;
```
brianhatchl, line 167: Couldn't we move these two lines outside the if/then so they aren't duplicated?

brianhatchl (changes requested):

modules/Hoot/ui/managePanel/reviewBookmarks/reviewBookmarkNotes.js
bobi7: the call `Hoot.layers.removeAllLoadedLayers();` will remove the value of `Hoot.ui.conflicts.data.forcedReviewItem` so order matters

brianhatchl (approved):

PR #1839, opened 2020-11-19
Fixed #1836. If custom extent is on map, default that for grail pull

### LINKED ISSUE #1836: if custom map data loaded, set custom extent as default for grail pull

bobi7:
ref #1836


brianhatchl (approved):
One day we should consolidate all those duplicate strings floating around into `const`

PR #1846, opened 2020-12-15
Fixed #1837. added ability to change overpass query for bobby magic

### LINKED ISSUE #1837: add ability to customize overpass query in TM panel
after user selects derive type can do popup box to allow overpass query select

bobi7:
ref #1837


brianhatchl (approved):
LGTM :owl:

PR #1854, opened 2021-01-15
Fixed #1852. Made layerAdd.js event listeners have unique identifier

### LINKED ISSUE #1852: imported datasets unable to be added to map until after refresh
after importing a dataset, if you go to the map and try to add reference you will see the dataset isn't in the list.
if you refresh the page it will finally show up there

bobi7:
ref #1852

### DIFF
--- a/modules/Hoot/ui/sidebar/layerAdd.js
+++ b/modules/Hoot/ui/sidebar/layerAdd.js
@@ -215,7 +215,8 @@ export default class LayerAdd extends SidebarForm {
      * Listen for events
      */
     listen() {
-        const className = this.constructor.name;
+        // Need refType to differentiate between panel for add reference dataset and add secondary dataset
+        const className = `${ this.constructor.name }_${ this.formMeta.refType }`;
 
         Hoot.events.listen( className, 'render-dataset-table', () => {
             this.renderFolderTree();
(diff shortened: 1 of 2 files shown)


modules/Hoot/ui/sidebar/layerAdd.js
```diff
@@ -215,7 +215,8 @@ export default class LayerAdd extends SidebarForm {
      * Listen for events
      */
     listen() {
-        const className = this.constructor.name;
+        // Need refType to differentiate between panel for add reference dataset and add secondary dataset
+        const className = `${ this.constructor.name }_${ this.formMeta.refType }`;
```
bobi7, line 219: layerAdd was called twice. once for add reference dataset panel and another for add secondary dataset. the issue is with the old implementation I was removing the event listeners from the reference and replacing it with the secondary panel since they shared the same class name and event name combo

brianhatchl: So before this fix, would the new dataset appear in the list for the secondary panel?

bobi7: correct

brianhatchl (approved):
:shipit:

PR #1861, opened 2021-02-19
Fixed #1860. grail pull will now load whatever dataset(s) were pulled

### LINKED ISSUE #1860: Grail pull should add reference dataset even when secondary dataset already present
This seems to work when Secondary Data Extent is used, but not when Custom Data Extent is used.

bobi7:
ref #1860


brianhatchl (approved):
LGTM :bow_and_arrow: :dart:

PR #1867, opened 2021-03-30
Fixed #1866. fixed error with running saved favorite conflation type

### LINKED ISSUE #1866: error with running saved favorite conflation type
after saving favorite options and trying to use that save set to run conflate errors occur

VGI-2324

bobi7:
ref #1866

### DIFF
--- a/modules/Hoot/ui/sidebar/advancedOpts.js
+++ b/modules/Hoot/ui/sidebar/advancedOpts.js
@@ -848,34 +848,30 @@ export default class AdvancedOpts {
         }
 
         function checkType( member ) {
+            let defaultVal;
             if ( member.input === 'checkbox' ) {
-                updateOpts.push(
-                    {
-                        input: member.input,
-                        default: d3.select(`#${member.id}`).select('input').property('checked'),
-                        id: member.id,
-                        description: member.description,
-                        label: member.label,
-                        type: member.type,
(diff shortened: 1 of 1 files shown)


modules/Hoot/ui/sidebar/advancedOpts.js
```diff
@@ -848,34 +848,30 @@ export default class AdvancedOpts {
...
-                        default: d3.select(`#${member.id}`).select('input').property('checked'),
-                        id: member.id,
-                        description: member.description,
-                        label: member.label,
-                        type: member.type,
-                        option: member.option
-
-                    }
-                );
+                defaultVal = d3.select(`#${member.id}`).select('input').property('checked');
```
maxgrossman: if the only thing that is different here is 'checked' versus 'value' could you just do a `defaultVal  = ${all that d3 stuff}.property(member.input === checkbox ? 'checked' : 'value')`

maxgrossman: even better just assign the value in the dictionary `selectedOpt` rather than creating the variable for it

bobi7: great idea

maxgrossman (approved):
tested and works as described. nice

Issue #1868, opened 2021-04-19
have translations be a folder tree like datasets


bobi7:
VGI 2324

PR #1869, opened 2021-04-19
Fixed 1868. added folder tree for translations

bobi7:
ref #1868


bobi7:
- [x] less of a gap between the buttons at the top
- [x] add public/private folders toggle button
- [x] change 'create folder' to 'add folder' to better match with the datasets panel

brianhatchl:
It seems attempting to move a folder the user does not own needs to populate the error alert.  The http response shows "You must own both folders to move it" and the alert popup appears but it is empty.

brianhatchl:
I see one more place where error message needs to be alerted.  When moving a translation script user does not own to a folder they do own.

brianhatchl:
I noticed that there is a context menu for Moving/Renaming a translation script but the popup only allows it to be moved.  The View translation option opens a form that lets description and the script to be edited, but does not allow the name or the path to be edited.  Maybe these could be combined to allow moving/renaming/editing in the view form or at least made consistent.

brianhatchl:
I also noticed a weirdness when moving a translation owned by my user from a public folder to a private folder, the highlighting remained blue (indicating public).  I think it should inherit the visibility of it's parent folder.

bobi7:
fixed the inherit parent folder visibility in services code

brianhatchl (approved):
LGTM :soccer: :trophy:

PR #1877, opened 2021-05-18
Fixed #1876. Fixed issue with dropdown in derive diff not populating

### LINKED ISSUE #1876: Derive Differential Changeset submit doesnt work

bobi7:
ref #1876 

related to: https://github.com/ngageoint/hootenanny/pull/4800

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1889,9 +1889,10 @@ div.combobox {
 .combobox-caret {
     display: inline-block;
     position: relative;
+    height: 30px;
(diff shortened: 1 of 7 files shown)


brianhatchl:
Think you could make the combobox the full width and then add the chosen options below the combobox (so scrolling to the right is not necessary)?

image

Or is this just how this control works?

brianhatchl (approved):
:rocket: :ringed_planet:

PR #1895, opened 2021-08-11
fixed #1894. fixed issue with how bounds for multi poly was drawn

### LINKED ISSUE #1894: Hootenanny/Grail -  Use precise secondary data pulls for POI to poly and POI to POI conflation
fix jobs panel information display when long polygon coordinate. limit it with ellipsis and on hove show full text

bobi7:
if 2 polygons were part of the bounds, a line would be drawn connecting the 2 polys

ref #1894

### DIFF
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -34,7 +34,7 @@ import {
 } from '../../util';
 
 import {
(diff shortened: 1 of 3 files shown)


brianhatchl (approved):
Noice :man_dancing:

PR #1897, opened 2021-08-11
Fixed #1896. enable diff conflate option for diff w tags. poly bug fix

### LINKED ISSUE #1896: Enable snapping option with Differential wTags conflation type
As a user I'd like to run differential w/tag and utilize hoots snapping capability when uploading the changeset.

Currently, when differential w/tags is elected as conflation type, the "differential" adv option becomes greyed out, and user can't select the pulldown bar to enable the auto-snapping process.

bobi7:
ref #1896 

poly bug fix for #1894

### DIFF
--- a/modules/Hoot/managers/layerManager.js
+++ b/modules/Hoot/managers/layerManager.js
@@ -36,6 +36,7 @@ import {
 import {
     polyStringToCoords, polyStringToCoordsList
 } from '../tools/utilities';
+import _cloneDeep from 'lodash-es/cloneDeep';
 
 export default class Layers {
     constructor( hoot ) {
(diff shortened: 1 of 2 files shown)


modules/Hoot/managers/layerManager.js
```diff
@@ -310,6 +311,8 @@ export default class Layers {
             let coordsList;
             if ( lyr.bounds && lyr.bounds.includes( ';' ) ) {
                 coordsList = polyStringToCoordsList( lyr.bounds );
+            } else {
+                coordsList = _cloneDeep( [ polyCoords ] );
```
brianhatchl, line 315: what was this fixing?  a single polygon vs multiple?

bobi7: yea the else is for single

brianhatchl (approved):
:checkered_flag: :racing_car:

PR #1901, opened 2021-08-19
Fixed #1900. fixed issue with import multiple not able to take prj file

### LINKED ISSUE #1900: prj files cant be imported on import multiple
When doing import multiple shapefiles, the .prj file gets greyed out and so the import fails

bobi7:
code cleanup added

ref #1900

### DIFF
--- a/modules/Hoot/ui/modals/importDataset.js
+++ b/modules/Hoot/ui/modals/importDataset.js
@@ -9,13 +9,11 @@ import _find    from 'lodash-es/find';
 import _forEach from 'lodash-es/forEach';
 import _reject  from 'lodash-es/reject';
 
-import FormFactory        from '../../tools/formFactory';
-import { getBrowserInfo } from '../../tools/utilities';
-import {
-    importSingleForm,
-}           from '../../config/domMetadata';
+import FormFactory from '../../tools/formFactory';
+import { getBrowserInfo, unallowableWordsExist } from '../../tools/utilities';
+import { importSingleForm } from '../../config/domMetadata';
 import _get from 'lodash-es/get';
-import axios         from 'axios/dist/axios';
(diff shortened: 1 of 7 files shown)


modules/Hoot/ui/modals/importDataset.js
```diff
@@ -71,6 +69,17 @@ export default class ImportDataset {
         ];
     }
 
+    static importTypeMap() {
```
bobi7, line 72: chose to add here because import multiple is using the same api logic as import single anyways so figure this would be the root

modules/Hoot/ui/modals/importDataset.js
```diff
@@ -71,6 +69,17 @@ export default class ImportDataset {
...
+        return {
+            'DIR': null,
+            'GEONAMES': '.geonames, .txt',
+            'OSM': '.osm, .osm.zip, .pbf',
+            'FILE': '.shp, .shx, .dbf, .prj, .zip',
+            'GEOJSON': '.geojson, .geo',
```
brianhatchl: is .geo really a valid extension for geojson?

bobi7: it was already there before so i assume someone put it for a good reason

google says "A GEO file is a 3D CAD file saved in the GEO format"

brianhatchl: I'd rather it be .json, can you make that switch?

brianhatchl: .geojson and .json

bobi7: fixed

brianhatchl (approved):
Nice refactoring :taco:

PR #1906, opened 2021-08-30
Fixed #1905. enable TM Panel to work with custom dataset

### LINKED ISSUE #1905: allow tasking manager panel to use imported dataset as secondary data

bobi7:
ref #1905


brianhatchl (approved):
Ooh la la :man_cook: :kiss:

PR #1914, opened 2021-10-25
Rate limit import multiple datasets and dataset delete requests

brianhatchl:
Fixes bug where multiple import requests could exhaust available db connections

Also allow zipped geojson files to be imported


brianhatchl:
I'm not sure why the tests are failing with
```
> iD@2.12.1 test:phantom /home/vagrant/hoot-ui
> phantomjs --web-security=no node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html spec
ReferenceError: Can't find variable: iD
   in global code at file:///home/vagrant/hoot-ui/test/spec/spec_helpers.js:4
```

I can only imagine that is has something to do with the introduction of `p-limit` as a first class dependency that then changed the versions used on transitive dependencies?

brianhatchl:
Here is a zip of 50 copies of a single point building geojson that can be uploaded to test.  The rate limit will do 20 concurrently.
[PAL015_files.zip](https://github.com/ngageoint/hootenanny-ui/files/7539726/PAL015_files.zip)

PR #1957, opened 2022-03-09
Make sure the context menu does not dip below the screen horizon

### LINKED ISSUE #1951: Dataset context menu should open up when at the bottom of the screen
so that the options are not hidden below the horizon

brianhatchl:
fixes #1951


maxgrossman (approved):
nice!

PR #1958, opened 2022-03-09
Need to track mapId when splitting ways

### LINKED ISSUE #1913: Marker css class for merged, reference, secondary is dropped when ways are split
When saved and reloaded/redrawn the proper rendering is restored.

brianhatchl:
fixes #1913


brianhatchl:
If you load DcGis and DcOsm and conflate and then split one of the ways (either merged or from ref or sec datasets) the datasource color should be preserved.  Before this fix the new way would be colored gray.

maxgrossman (approved):
nice!

PR #1974, opened 2022-05-27
make translation folder path requirement consistent

### LINKED ISSUE #1973: Translations dialog has inconsistent enforcement of folder path
The Add Translation dialog allows saving to root, but the Modify Translation dialog enforces that a folder path be set.

brianhatchl:
between add and modify
fixes #1973

### DIFF
--- a/modules/Hoot/managers/folderManager.js
+++ b/modules/Hoot/managers/folderManager.js
@@ -295,7 +295,16 @@ export default class FolderManager {
     }
 
     get translationFolders() {
-        return this._translationFolders.filter(this.filterTranslationVisible);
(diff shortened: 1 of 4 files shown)


brianhatchl:
This change allows saving a custom translation to the root folder.  This differs from the treatment of datasets, but translations are far less numerous so we'll keep it this way for the time being.

brianhatchl:
From Max,
>On review, I can add files to nested folders, but i cannot move them back to root once they’re in a non root

maybe it just needs a "/" in the translations folder list.

maxgrossman (approved):

Issue #1976, opened 2022-06-16
Advanced conflation options pane orphaned if layers removed


brianhatchl:
If adv opts is open when layers are removed it is orphaned on the screen and unable to be closed/hidden.

PR #1987, opened 2022-08-25
update adv opts panel to toggle if open when layer deleted

jackgrossman18:
when one of the two layers are deleted while adv opts panel is open, panel will toggle to disappear

ref #1976


brianhatchl (approved):
:dark_sunglasses: at :night_with_stars:

Issue #1977, opened 2022-06-16
Conflate button colored disk stacks should update when current ref/sec color selection changes


brianhatchl:
![image](https://user-images.githubusercontent.com/2600932/174094428-f60aa1e6-cb43-4da8-b0f1-bf21384007cd.png)

PR #1988, opened 2022-08-26
update colors shown in conflate section when layer colors changed

jackgrossman18:
ref #1977

### DIFF
--- a/modules/Hoot/ui/sidebar/layerConflate.js
+++ b/modules/Hoot/ui/sidebar/layerConflate.js
@@ -72,8 +72,10 @@ class LayerConflate extends SidebarForm {
     }
 
     updateLayerColors(newColor, refType) {
-        this.fieldset.selectAll(`.${refType}.thumb`)
-            .attr('class', `${refType} thumb round _icon data light contain inline fill-${ newColor }` );
+        let thumb = this.fieldset.select(`.${refType}.thumb`);
+        let oldColor = thumb.attr('class').split(' ').find(d => d.startsWith('fill-'));
+        thumb.classed(`${ oldColor }`, false);
+        thumb.classed(`fill-${ newColor }`, true);
     }
 
     checkForFavorite() {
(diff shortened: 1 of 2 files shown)


modules/Hoot/ui/sidebar/layerConflate.js
```diff
@@ -106,7 +111,8 @@ class LayerConflate extends SidebarForm {
             .selectAll( '.thumb' )
             .data( layers ).enter()
             .append( 'div' )
-            .attr( 'class', d => `thumb round _icon data light contain inline fill-${ d.color }` );
+            .attr( 'class', d => `thumb round _icon data light contain inline fill-${ d.color }` )
+            .attr( 'id', d => `${d.refType}` );
```
brianhatchl: the sidebar layer form already uses this `id`.  might be better to put it in the `class`.
```
.attr( 'class', d => `${d.refType} thumb round _icon data light contain inline fill-${ d.color }` );
```

modules/Hoot/ui/sidebar/layerConflate.js
```diff
@@ -71,6 +71,11 @@ class LayerConflate extends SidebarForm {
         this.fieldset = new FormFactory().createFieldSets( this.innerWrapper, this.formData );
     }
 
+    updateLayerColors(newColor, refType) {
+        this.fieldset.selectAll(`#${refType}.thumb`)
+            .attr('class', `thumb round _icon data light contain inline fill-${ newColor }` );
```
brianhatchl: This kind of approach is a bit more wordy, but will avoid duplicating the entire `class` list for the thumbnail.
```
        let thumb = this.fieldset.select(`.${refType}.thumb`);
        let oldColor = thumb.attr('class').split(' ').find(d => d.startsWith('fill-'));
        thumb.classed(`${ oldColor }`, false);
        thumb.classed(`fill-${ newColor }`, true);
```

brianhatchl (changes requested):
It works great but I have a couple implementation suggestions.

brianhatchl (approved):
:dark_sunglasses: :safety_vest: :headphones: :video_game:
yeyeyeyeye

Issue #1978, opened 2022-06-16
Add new folder not working from conflate form


brianhatchl:
I think since writing conflate output to the root `/` was disallowed, the create new folder option may be broken.  (The folder drop down does not allow the root selection anymore).

PR #1989, opened 2022-08-29
remove add new folder option from conflate form

jackgrossman18:
ref #1978

### DIFF
--- a/modules/Hoot/ui/sidebar/layerConflate.js
+++ b/modules/Hoot/ui/sidebar/layerConflate.js
@@ -284,7 +284,7 @@ class LayerConflate extends SidebarForm {
         Hoot.layers.loadLayer( params );
     }
 
-    handleSubmit() {
+    async handleSubmit() {
         d3.event.stopPropagation();
         d3.event.preventDefault();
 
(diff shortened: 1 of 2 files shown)


modules/Hoot/ui/sidebar/layerConflate.js
```diff
@@ -300,6 +300,13 @@ class LayerConflate extends SidebarForm {
             this.advancedOptions.toggle();
         }
 
+        if (this.newFolderNameInput.node().value.length > 0) {
+            let newFolderName = this.newFolderNameInput.node().value;
+            let newFolderId;
+            newFolderId = (await Hoot.folders.addFolder( newFolderName, newFolderName )).folderId;
```
brianhatchl: I think if we want the new folder to always be created at the root, then the first param to `addFolder` should be empty string.  This sort of works if the `newFolderName` doesn't exist, but would get weird if the folder name did exist and would create another folder with the same name underneath.

It might be a little confusing that the `Path` drop down doesn't allow a root ` / ` folder to be shown, even though the new folder will be created at the root.  Maybe the placeholder for the `Path` drop down should be ` / `, but if not creating a new folder it should require the user to make a choice of an existing folder.

modules/Hoot/ui/sidebar/layerConflate.js
```diff
@@ -300,6 +300,13 @@ class LayerConflate extends SidebarForm {
             this.advancedOptions.toggle();
         }
 
+        if (this.newFolderNameInput.node().value.length > 0) {
+            let newFolderName = this.newFolderNameInput.node().value;
```
brianhatchl: I prefer the `.property('value')` d3 syntax over `.node().value`

brianhatchl (changes requested):

brianhatchl (approved):
The current behavior to use the folder of either the Ref or Sec layer dataset as the default output folder for the conflation result makes the most sense.
Because either of those layer must already have a folder, and it's logical that a merged output would go to the same folder, we don't really need the New Folder Name text box here at all

modules/Hoot/ui/sidebar/layerConflate.js
```diff
@@ -284,7 +283,7 @@ class LayerConflate extends SidebarForm {
         Hoot.layers.loadLayer( params );
     }
 
-    handleSubmit() {
+    async handleSubmit() {
```
brianhatchl: actually, can revert this change

PR #1990, opened 2022-08-30
fix js error when adv opt form not shown

brianhatchl:
Found this error when trying to remove a single layer from the map
```
advancedOpts.js:41 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'classed')
    at get isOpen [as isOpen] (advancedOpts.js:41:1)
    at Layers.removeLoadedLayer (layerManager.js:492:1)
    at HTMLButtonElement.<anonymous> (sidebarController.js:175:1)
```


jackgrossman18 (approved):
🕶️ 👍

PR #1994, opened 2022-09-06
address dependabot/terser problem by adding terser as dep

maxgrossman:
i'm not sure why the editor-layer-index change happened but the terser changes seem identical to what's in the pr from dependabot


brianhatchl:
I pushed a version with just the terser update #1999 
It did end up grabbing terser `5.15.0`

PR #1998, opened 2022-09-07
allow for variable overrides via globals script

maxgrossman:
add a `globals` file that can be use to override external api's locations with if they differ from default.
doing overrides in this separate file allows for flexibility to make changes after an rpm is installed by just making changes to a single file.

For example if my the path to my translation server is different then you'd update the globals file to be.

```js
var tm4ApiUrl,
    translationServerPort = 'somewhere_else',
    mergeServerPort;
```


dist/globals
```diff
@@ -0,0 +1,3 @@
+var tm4ApiUrl,
+    translationServerPort,
+    mergeServerPort;
```
brianhatchl: any reason not to name this one `globals.js`?

maxgrossman: this npm script removes all dist/*.js files and i was confused when i kept deleting `globals.js` unknowingly
https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/package.json#L17

maxgrossman: i'm all for a better solution if we have one.

brianhatchl: maybe change the rm to `dist/*.min.js`?

maxgrossman: 2️⃣ ⌚ 🔙 2️⃣ 🔙

brianhatchl (approved):
:alarm_clock: up & :lock: it in

PR #2001, opened 2022-09-12
Enable double-click add to map in layer liist

### LINKED ISSUE #1991: Add a double-click handler for Reference and Secondary layer list entries
double-click should add the layer to the map, like pressing the `Add Layer` button does.

brianhatchl:
fixes #1991

### DIFF
--- a/modules/Hoot/tools/folderTree.js
+++ b/modules/Hoot/tools/folderTree.js
@@ -341,6 +341,13 @@ export default class FolderTree extends EventEmitter {
                 // use self as context but still pass in the clicked element
                 self.click.call( self, this, d );
             } )
+            .on( 'dblclick', d => {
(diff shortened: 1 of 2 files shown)


modules/Hoot/tools/folderTree.js
```diff
@@ -909,4 +916,9 @@ export default class FolderTree extends EventEmitter {
 
         this.update( d );
     }
+
+    setDoubleClickHandler(ref, handler) {
+        this.doubleClickRef = ref;
+        this.doubleClickHandler = handler;
```
brianhatchl, line 922: this feels kinda hacky, but not sure how else to call `submitLayer` from `folderTree`

maxgrossman (approved):

PR #2008, opened 2022-09-23
add export option for thematic groups

### LINKED ISSUE #2005: Add an export UI element to override "ogr.thematic.structure"
>=== ogr.thematic.structure
>* Data Type: bool
>* Default Value: `true`
>
>Where applicable, export data in Thematic Groups (TransportationGroundCrv, StructurePnt etc)
instead of one FCODE per file/layer (ROAD_L, BUILDING_P etc).

brianhatchl:
fixes #2005

### DIFF
--- a/modules/Hoot/config/domMetadata.js
+++ b/modules/Hoot/config/domMetadata.js
@@ -387,6 +387,7 @@ export function exportDataForm( zipOutput ) {
           exportFormatId = 'exportFormatCombo',
           exportNameId = 'dataExportNameTextInput',
           exportFgdbId = 'exportAppendFgdb',
+          exportThematic = 'exportThematic',
           exportHootTags = 'exportHootTags',
           that = this;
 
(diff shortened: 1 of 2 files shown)


modules/Hoot/config/domMetadata.js
```diff
@@ -397,6 +398,12 @@ export function exportDataForm( zipOutput ) {
         d3.select( `#${exportFgdbId}_container` )
             .classed( 'hidden', !showFgdbTemplate );
 
+        const showThematic = d3.select( `#${exportFormatId}` ).property( 'value' ).indexOf('OpenStreetMap') === -1
+                        && d3.select( `#${exportComboId}` ).property( 'value' ).indexOf('OSM') === -1;
```
brianhatchl, line 402: Show option to suppress thematic grouping of layers only when both format and schema are not osm.

PR #2010, opened 2022-10-03
Add feature derive changeset as JOSM osm with action attributes

brianhatchl:
Adds conditional for Open In JOSM dataset context menu
Adds new action for conflated grail layers in Jobs panel


brianhatchl:
For the Datasets Panel, after highlighting and righ-clicking a dataset, the context menu has an `Open Dataset in JOSM` option.

For imported datasets and non-grail merged datasets, this link will run an Export job and then use remote control to load that export file in JOSM.

The newness in this PR is that for grail merged datasets, the link will run a Derive Changeset job with the output format being JOSM osm with action attributes on the elements.  Remote control should load this export file and immediately have the Upload button enabled with all the changes from the conflation (plus any review relations if those are present).


Screenshot from 2022-10-04 13-23-04
There is also a new action icon in the Jobs Panel for grail conflate jobs to `derive josm .osm`

PR #2019, opened 2022-11-17
Enhance zoom to edit notice

brianhatchl:
It will now reflect when a layer needs to be added first.

### DIFF
--- a/modules/ui/notice.js
+++ b/modules/ui/notice.js
@@ -17,9 +17,6 @@ export function uiNotice(context) {
         var button = div
             .append('button')
             .attr('class', 'zoom-to notice fillD')
-            .on('click', function() {
-                context.map().zoomEase(context.minEditableZoom());
-            })
             .on('wheel', function() {   // let wheel events pass through #4482
                 var e2 = new WheelEvent(d3_event.type, d3_event);
                 context.surface().node().dispatchEvent(e2);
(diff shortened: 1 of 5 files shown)


brianhatchl:
needs to not zoom, but expand add layer when notice of add layer to edit is clicked

brianhatchl:
also, zoom to edit apparently does not zoom, so fix that too

brianhatchl:
I think the issues above have all been addressed

modules/ui/notice.js
```diff
@@ -32,9 +30,20 @@ export function uiNotice(context) {
...
+            var noLayers = _isEmpty(Hoot.layers.loadedLayers);
+            var tooHigh = context.map().zoom() < context.minEditableZoom();
+            var tooManyNodes = context.map().tooManyNodes();
+            var canEdit = !tooHigh && !tooManyNodes && !noLayers;
             div.style('display', canEdit ? 'none' : 'block');
+            button.on('click', function() {
+                if (noLayers) {
+                    context.map().addDataset();
+                } else {
+                    context.map().zoomEase(16); //upstream min editable zoom
```
maxgrossman, line 42: Just curious, i see this moves away from zooming to `context.minEditableZoom()` and just classic zoom 16.
I think that just ends up changing us back to using 16 as opposed to [9](https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/core/context.js#L178)?

brianhatchl: In order to display sparse data a lower zooms we changed the min editable zoom for hoot-ui to be 9.  I think rather than zoom to 9 (which may still have too many nodes) I chose to zoom to the O.G. z16

maxgrossman (approved):

Issue #1437, opened 2019-05-22
use translation server in translation assistant


maxgrossman:
Not something we need/should put efforts towards now, but we can probably just use the translation server for translation assistant. This could remove the need for the 300kb tds mappings file the class currently uses

brianhatchl:
Yes, these are under the `/data` directory and are called `*_field_values.josn`

brianhatchl:
Figure we can use taginfo to replace `osm-plus-taginfo.json`.
Requests could be to fetch a page of keys/values by popularity (count), e.g. 

>Get popular keys
https://taginfo.openstreetmap.org/api/4/keys/all?include=prevalent_values&sortname=count_all&sortorder=desc&page=1&rp=18&qtype=key

>Get popular values for "building"
https://taginfo.openstreetmap.org/api/4/key/values?key=building&filter=all&lang=en&sortname=count&sortorder=desc&page=1&rp=16&qtype=value


or in the taginfo api with a `query` of the user typed text, e.g.

>Get keys that include "aerow"
https://taginfo.openstreetmap.org/api/4/keys/all?lang=en&page=1&query=aerow&rp=10&sortname=count&sortorder=desc

>Get values for "aeroway" that include "runw"
https://taginfo.openstreetmap.org/api/4/key/values?filter=ways&key=aeroway&lang=en&page=1&query=runw&rp=25&sortname=count&sortorder=desc

brianhatchl:
fixed in #2032

PR #2032, opened 2023-01-30
updating ui to handle new translation server endpoints

jackgrossman18:


brianhatchl (changes requested):
It's looking good, but we'll need to do this https://github.com/ngageoint/hootenanny-ui/issues/1437#issuecomment-1406650444 for the OSM schema because it's not supported by translation server.

You can also remove the OBE translation *_field_values.json files.

brianhatchl:
Looking into this further, there is already a class `taginfo` that will make the necessary `keys` requests.  I noticed this when looking at the network traffic while using TA.  The request in the class uses `100` for the `rp` param which is record count.  I think we'll need that many to offer a substantial enough list of possibilities.  There are some examples of it's use in `raw_tag_editor.js`.  There is also a `values` call in taginfo.  Re-using this taginfo class makes sense.  But also modify it to accept an `apibase` override from `globals.js`

I like the change to pull the available schemas from Translation Server.  Can you change the options buttons to a drop down though?

modules/Hoot/ui/managePanel/transAssist/upload.js
```diff
@@ -7,26 +7,16 @@
...
-                name: 'TDSv70',
-                enabled: true
-            },
-            {
-                name: 'TDSv61',
-                enabled: true
-            },
-            {
-                name: 'MGCP',
+        this.schemaOptions = Hoot.translations.availableTranslations.map(v => {
```
brianhatchl: can you make `schemaOpts` in this class a drop down instead of a list of radio buttons?

modules/Hoot/ui/managePanel/transAssist/tagMapWidget.js
```diff
@@ -29,20 +29,38 @@ export default class TagMapWidget {
...
+        if (!Hoot.config.tagInfo[this.schemaOption]) {
+            Hoot.config.tagInfo[this.schemaOption] = [];
+        }
+        if (Hoot.config.tagInfo[this.schemaOption].length === 0) {
+            (this.schemaOption === 'OSM'
+                ?  new Promise(function(resolve, reject) {
+                    services.taginfo.keys({}, function(err, data) {
+                        if (!err) {resolve(data.map(v => { return v.value; }));}
+                        else { reject(err);}
+                    });
```
brianhatchl: I think this call needs to move to the `change()` function so that it can pass a query param.  For example, I type the tag `amenity` but it must not be in the top 10 most popular tag keys.  When doing this the manual filtering in `change()` won't be needed for taginfo results.

brianhatchl (changes requested):
In addition to the inline comments, there is also a js error when pressing enter on a tag key that's not in taginfo.  This should still be allowed but just won't reveal any tag values.

modules/Hoot/ui/managePanel/transAssist/tagMapWidget.js
maxgrossman: with changes in [1696e58](https://github.com/ngageoint/hootenanny-ui/pull/2032/commits/1696e58ac0c5af87c3164a7a98480f16ab30d549) think we also want the if block on line 51 to be `is tagInfo list empty and is schema not osm` then it should only invoke the `getFieldMappings` func.

modules/Hoot/ui/managePanel/transAssist/upload.js
```diff
@@ -3,28 +3,32 @@
...
-                enabled: true,
-                checked: true
-            },
-            {
-                name: 'TDSv70',
-                enabled: true
-            },
-            {
-                name: 'TDSv61',
+        this.schemaOptions = Hoot.translations.availableTranslations.map(v => {
```
maxgrossman: as combobox maybe the mapping here is just to a list of objects like `{key: value, value: value}`. I think enabled was a concept tied to when these were radio inputs.

modules/Hoot/ui/managePanel/transAssist/upload.js
```diff
@@ -51,46 +55,75 @@ export default class Upload {
...
         this.createUploadButtons();
+        this.loadSchemas();
     }
 
     createUploadForm() {
         this.uploadForm = this.instance.panelWrapper
             .append( 'form' )
             .classed( 'ta-upload-form round keyline-all fill-white', true );
     }
+    async loadSchemas() {
```
maxgrossman: I don't think we need this function as it just swaps this.schemaOptions over to this.schemaList.

modules/Hoot/ui/managePanel/transAssist/upload.js
```diff
@@ -51,46 +55,75 @@ export default class Upload {
...
+            this.schemaList = this.schemaOptions;
+        } catch ( err ) {
+            err.message = err.data;
+            Hoot.message.alert( err );
+        } finally {
+            this.populateFilterCombos();
+        }
+    }
+
+    populateFilterCombos() {
```
maxgrossman: I think this function can be simplified to just append the '#tagSchema` input to the `ta-filter-container`. Then you can use the `call` function on that input and pass the d3combobox to it to 'comboboxify' it. better yet if we make suggested changes to schemaOptions you can just pass schemaOptions.

maxgrossman: This could even be achieved right inside createSchemaSelector, letting us get rid of the populateFIlterCombos function

brianhatchl (approved):
Ok, the combobox filtering issue I saw turned out to be related to the dev console being open and the vertical screen real estate not being large enough for the combobox to render when dropping below the input control.  So the options were filtering but the drop down was rendering as like a 1px div.

Closing the dev console allowed it to work, at least for the first attribute mapping.

PR #2034, opened 2023-02-01
Translated tagging fixes

maxgrossman:
- make raw tag editor ignore translated keys for fcode that are already in tags returned from translation server
- make raw tag editor show descriptions for keys and enumerated values when viewing translated features
- only append 'universal presets' when schema is osm

### DIFF
--- a/modules/ui/raw_tag_editor.js
+++ b/modules/ui/raw_tag_editor.js
@@ -206,12 +206,17 @@ export function uiRawTagEditor(context) {
             });
 
         items.selectAll('input.key')
-            .attr('title', function(d) { return d.key; })
+            .attr('title', function(d) { return Hoot.translations.decodeSchemaKey(d.key); })
             .call(utilGetSetValue, function(d) { return d.key; })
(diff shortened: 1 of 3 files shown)


modules/ui/raw_tag_editor.js
```diff
@@ -206,12 +206,17 @@ export function uiRawTagEditor(context) {
...
             .property('disabled', isReadOnly);
 
         items.selectAll('input.value')
-            .attr('title', function(d) { return d.value; })
+            .attr('title', function(d) {
+                // when multi-value, show new line delimited so you can see each value
+                // when translated w/enumerated value, get title of enumerated value
+                // all other cases what's in the input has all the info user needs.
+                return Array.isArray(d.value) ? d.value.filter(Boolean).join('\n') : Hoot.translations.decodeSchemaValue(d);
+            })
```
brianhatchl, line 219: What's an example of a multi-valued tag?  I'm trying to think of how to test this.

brianhatchl (approved):
Question below, but LGTM :100:

PR #2037, opened 2023-02-08
Add show reviews operation

maxgrossman:
show_review_relations

not shown hear is that if you click the selected feature it will turn off the review relation svgs.


modules/svg/hoot.js
```diff
@@ -46,14 +85,64 @@ export function svgHoot(projection, context, dispatch) {
             .append('path')
             .merge(paths)
             .attr('class', function(d) {
-                if (!(_isEmpty(d))) {
```
maxgrossman, line 49: maybe i removed this without fully understanding it's original purpose -> https://github.com/ngageoint/hootenanny-ui/commit/4cf5295f12d3fe6cd7a2b7c5bc2bddce91a0e043

i have the question, when would the data element ever be empty?

modules/svg/hoot.js
```diff
@@ -81,11 +170,7 @@ export function svgHoot(projection, context, dispatch) {
     drawHoot.geojson = function(gj) {
         if (!arguments.length) return svgHoot.geojson;
 
-        if ( _isArray( gj ) ) {
-            svgHoot.geojson = svgHoot.geojson.concat( gj );
-        } else {
-            svgHoot.geojson.push( gj );
-        }
+        svgHoot.geojson = gj;
```
maxgrossman: when i was trying to see where we called this geojson() function i found once case in [layerManager](https://github.com/ngageoint/hootenanny-ui/blob/hoot2x/modules/Hoot/managers/layerManager.js#L332-L339).

I used that to determine 'we should make the geojson here a feature collection'. maybe instead it should be 'we should just keep an array' and change the linked invocation to take an array.

brianhatchl (approved):
:surfer: for sure!

PR #2039, opened 2023-02-09
update schema switcher, match id

jackgrossman18:
updating schema switcher to reflect functionality of id


css/80_app.css
```diff
@@ -1039,6 +1039,7 @@ a.hide-toggle {
     left: 60px;
     border-left: 1px solid rgba(0, 0, 0, .1);
     border-radius: 0 3px 3px 0;
+    width: 75%;
```
brianhatchl, line 1042: Do we also need this?
```
.schema-loading {
    background-image: url(img/loading-spin-dark.svg);
    background-position: 0 0;
    position: absolute;
    width: 40px;
    height: 40px;
    display: inline-block;
    margin-left: 5%;
}
```
I see that css class being introduced below

modules/ui/preset_list.js
```diff
@@ -427,6 +427,29 @@ export function uiPresetList(context) {
...
+            } else {
+                Hoot.translations.addTagsForFcode(context.geometry(_entityID), preset, context, _entityID, function(p) {
+                    p.icon = preset.icon;
+                    context.presets().replace(p);
+                    choosePreset(p);
+                });
+            }
+
             context.presets().choose(preset);
             context.perform(
```
brianhatchl: Do we still need this if doing `context.peform()` in `choosePreset`?

brianhatchl:
I noticed some missing presets for new translation schemas so I will attempt to add those.

brianhatchl (approved):

Issue #1548, opened 2019-08-14
EVWHS plugin has multiple issues


brianhatchl:
1. The basemap entry with stacking profile options is not appearing when the imagery plugin is enabled.
2. The footprints collection overlay is not able to be turned off once checked and it seems to repeatedly repeat tiled requests in a loop.

PR #2040, opened 2023-02-14
Get dg services working

maxgrossman:
Addresses problems 1 and 2 in #1548. 

Additionally,
- adds entries in overlay list for images selected from carousel
- fixed issue with metadata popup not updating rendered metadata when I'd cycle through images.
- adds in opacity slider for carousel images


brianhatchl (approved):
slingin' :flying_disc:

PR #2042, opened 2023-03-06
add env var overrides for webpack conf

maxgrossman:
for https://github.com/ngageoint/hootenanny/pull/5583


brianhatchl (approved):
works with the docker pr branch and also works (doesn't break) running webpack in it's own repo talking to vagrant vm

Issue #1967, opened 2022-05-16
Post-conflation review (red/blue) table labels


MappinBenny:
It would be helpful while resolving post-conflation issues to add a label showing which dataset is which in the red/blue tables.
![hoot_screenshot](https://user-images.githubusercontent.com/102063985/168686217-6c55b005-02a5-4247-acb5-dc789fab3553.png)

brianhatchl:
Most times the red column is the reference and the blue column is the secondary.  But there are situations where a review may involve two features from the same layer or against a feature that has been merged during conflation.

How about something like this?

![Screenshot from 2022-06-03 12-44-27](https://user-images.githubusercontent.com/2600932/171909412-9f124046-e291-4b9a-9e98-55ca78abdf9e.png)

![Screenshot from 2022-06-03 12-44-10](https://user-images.githubusercontent.com/2600932/171909437-8072d5b9-8418-4224-8fdc-0a4e27fa8b1a.png)

![Screenshot from 2022-06-03 12-45-00](https://user-images.githubusercontent.com/2600932/171909452-e55c60e5-f97e-4a6c-b924-1489bfb9a97a.png)

MappinBenny:
Brian... I am so sorry and embarrassed to reply 10 months after this discussion started... your suggestion would work great to help with clarity during the review process!

PR #2044, opened 2023-03-09
Post conflation review table update

jackgrossman18:
ref #1967


maxgrossman (commented):
looks good to me.
tagging hatchl since he was original person thinking about solution here.

brianhatchl:
I think we need to check the `hoot:status` tag of the two features because it is mostly but not always Reference vs Secondary.
https://github.com/ngageoint/hootenanny-ui/issues/1967#issuecomment-1146168847

modules/Hoot/ui/conflicts/info.js
```diff
@@ -89,6 +89,9 @@ export default class ConflictMetadata {
                 .html( navHtml );
         }
 
+        // add reference and secondary layer title row
+        tagsMerged.unshift({key: '', value: ['Reference', 'Secondary']});
+
```
brianhatchl: I think this chunk can be removed now that `filterHootStatus` is there.

brianhatchl (changes requested):

brianhatchl (approved):

PR #2046, opened 2023-03-15
remove more files now that TA uses translation server

brianhatchl:
vs these static tag sample files

refs #1437


jackgrossman18 (approved):
LGTM 🍝

PR #2054, opened 2023-04-21
add text for empty stats data, update css to fit 'no stats' text

jackgrossman18:


brianhatchl (approved):
The "empty" table looked a bit wonky, so I just made it omit the Stats entry all together.

PR #2057, opened 2023-05-16
add check for nodes in entity when multipolygon

jackgrossman18:


modules/operations/review.js
```diff
@@ -9,7 +9,8 @@ import {
 
 export function operationReview(selectedIDs, context) {
     const entityId = selectedIDs[0];
-    const entityLoc = getLocation(context.hasEntity(entityId));
+    let members = context.hasEntity(entityId).members;
+    const entityLoc = members ?  getLocation(context.hasEntity(members[0].id)) : getLocation(context.hasEntity(entityId));
```
brianhatchl: I found another edge case where this code will have to filter out relations or traverse down relations if the first member of the relation is also a relation.  You can reproduce with the attached file.  Select the point feature and then select it's parent "site" relation.
[site_relation.zip](https://github.com/ngageoint/hootenanny-ui/files/11501155/site_relation.zip)

modules/operations/review.js
```diff
@@ -22,6 +23,17 @@ export function operationReview(selectedIDs, context) {
...
+    function getMemberNodes(members) {
+        if (members && members.length > 0) {
+            if (members[0].type === 'relation') {
+                let relationMembers = context.hasEntity(members[0].id).members;
+                return relationMembers[0];
+            } else {
+                return members[0];
+            }
+        }
+    }
```
brianhatchl: I think there could be situations where the relation has only relation members, so a while loop is in order.
How about something like:
```
    let entity = context.hasEntity(entityId);
    while (entity && entity.type === 'relation') {
        let members = entity.members.map(m => context.hasEntity(m.id));
        let nonRelationMember = members.find(e => e.type !== 'relation');
        if (nonRelationMember) {
            entity = nonRelationMember;
        } else {
            entity = members.find(m => true);
        }
    }

    const entityLoc = getLocation(entity);
```

brianhatchl (approved):
:owl:

PR #2059, opened 2023-06-12
if changeset stats has error property, display and make the upload view only

brianhatchl:
so the changeset can't be uploaded


brianhatchl:
If an changeset has `FCODE` or `F_CODE` tag keys on a feature, this error will show in the changeset stats popup and there won't be a submit button (similar to how the changeset stats popup appears for a stale changeset that has already been uploaded)
Screenshot from 2023-06-13 14-29-34


When a layer is added to the map and the first features are fetched and parsed a check for `FCODE` or `F_CODE` will run and display this error popup if detected
Screenshot from 2023-06-13 14-31-04

Issue #2071, opened 2023-11-16
Show the appendtoFgdb checkmark for all FGDB exports


mattjdnv:
The current behavior is to only show the Append to FGDB option when exporting to an FGDB with a translation that starts with "TDS".

Change this to show the export option for all FGDB exports.

mattjdnv:
Merged

PR #2072, opened 2023-11-16
Changed the appendToFGDB criterion

mattjdnv:
refs: #2071

Changed it so that it doesn't require a translation starting with "TDS"


abalosc1 (approved):
:rocket:

